PR #86, opened 2014-10-23
ping custom channels

fafafaf:
not pinging for #aeolus and other 'crucial' channels


HardlySoftly:
To clarify #aeolus was moved back to being a crucial channel.

Sheeo:
So we should be good to merge this?

I'm sorry it's been left hanging for so long.

IDragonfire:
Why this is on hold?
d7d7810f4adf86f8a8dea38c852c48e635090912

PR #159, opened 2015-01-03
Include GitHub/FAForever in links menu

HardlySoftly:
Included reference to Git in welcome message.

Removed spaces from before the exclamation marks in welcome message - feedback please if people don't want this.


HardlySoftly:
Also wondering if new link should be added to production.py, it seems to house a few already (Though I don't know what they're there for).

Sheeo:
:+1:

We've had several discussions about a central page that's better than "what's new" currently is. I'm not sure how to approach it, though.

Sheeo:
Merged with 155b739407e2ca4f15a9f398af78a7cfbfc9beae

Issue #146, opened 2014-11-27
Easy way to link to replay


Crotalus:
Make an easy way to link to replays in the lobby, for instance - transform #123782626 into a fafreplay-link when talking in a channel / with someone

PR #183, opened 2015-03-11
Convert #123456 -> link to replay vault

Crotalus:
A quick fix to convert #1234567 into replay links

Would be nice in the future if it could grab info about the replay and display in the tooltip.


Sheeo:
Indeed a bit quick.

I would prefer us to replace the ID's on reading, rather than replacing the before send. This means we don't burden IRC users with HTML, and allows the client to link to replays sent by IRC clients as well.

I'd like to have generic 'unfurling' of links, along with having this working for live games as well -- so you can go directly to live preview games (or potentially view a synced game with friends).

Crotalus:
AFAIK this is executed just before inserting a line into the chatArea, not sure why the function is called irc_escape. It's the same place as the other "linkifying" of hrefs. They are sent to IRC and back?

Sheeo:
Yes. `irc_escape` is a meaningful name, if it did only that

Sheeo:
We should track this on the issue for it here #146.

Also, I do think it's a lot better to do this when rendering the chat, so replay-id's sent by IRC users also get autolinked, and we don't pollute the IRC chat with HTML. We have more overall knowledge at the recipient than the sender, too -- i.e. you could imagine a user only configuring it to link certain replays.

Sheeo:
So QAI does this these days. Of course we should have a better feature for it embedded in the client.

We need functionality for it in the API first, though.

PR #198, opened 2015-04-04
Draft/model

IDragonfire:
This PR is only for discussion.

Added:
* Client Action API: https://github.com/IDragonfire/lobby/commit/21131807e0ffaf44931b2de750297dcd33af076f
* Admin Action API: https://github.com/IDragonfire/lobby/commit/f8f09ea72b1a891e32326b5e06f355832110e2bf
* Server Package API: https://github.com/IDragonfire/lobby/commit/a4c1069485062f3377d6ffb7a6ed93606579b98e

This reduce the LoC of the `_clientwindow` and more important decouple the architecture.


IDragonfire:
That is the first step into a MVC structure over an observer pattern (first with QtSignals, later maybe with pure python) ...

Currently every widget connects to a qt.signal (=server message), 
e.g. https://github.com/FAForever/client/blob/develop/src/client/_clientwindow.py#L1884

Their have their own internal model ...
Every widget must know the format of the message ...
and handle the different message types ....

####

This draft bundle all stuff into several classes:

 server_packets that currently translate server messages into Signals
(in the future it will update the Lobby Model, e.g. the Player Service) 

And the Client/Admin Actions to provide an temporary interface to modify the model.
(e.g. change Password, change UserStatus, change Avatar, ,,,)

The next step is to Create the Model and modify the lobby model.

###
Widget register to model events/data changes (not server messages) and use well defined method to modify the model.

The model validate the modification and if valid notify the server.
(e.g. validation must NOT be on client site, let ask the server, or to reduce the traffic use client side, or a shared component fromt the server)

PR #201, opened 2015-04-05
Limit the waitForInfo loop.

### LINKED ISSUE #187: forever.log growing indefinitely until disk full
Log is filled with this error

    2015-03-24 10:34:20,128 ERROR    faf.secondaryServer  Operation timed out while waiting for info.

vinzBad:
This fixes #187. Something on my machine garbled the diff..I'll fix this, but thought you could still give me some feedback?

Does this need to be rebased? If I understand rebasing correctly, this pull request already has a nice linear history going from the latest commit in develop.


Sheeo:
This should not need rebasing, rebases are only necessary iff commits were based off of a different commit than current develop/merge target.

I would like you to seperate the whitespace changes into a seperate commit though, it makes it easier to review and easier to work around when using git blame.

I don't know about this approach to solving the problem, we ought to be using the `self.serverSocket.readyRead` signal to read from the socket, which would eliminate the busywait.

vinzBad:
I also think that my approch is kinda ugly but this fits the theme of the module :P I'll take a look at ```self.serverSocket.readyRead```

Sheeo:
This will need rebasing onto current develop :)

PR #209, opened 2015-06-04
added a dropdown to sort the games list

HaraldWeber:
The games list is sometimes snappy when games are added or removed. So I added a drop down menu to sort games by player count, game quality and avg. player score in a game.

### DIFF
--- a/src/games/_gameswidget.py
+++ b/src/games/_gameswidget.py
@@ -133,6 +133,10 @@ class GamesWidget(FormClass, BaseClass):
 
         self.gameList.setItemDelegate(GameItemDelegate(self))
         self.gameList.itemDoubleClicked.connect(self.gameDoubleClicked)
+        self.gameList.sortBy = 0 # Default Sorting is By Players count
+
+        self.sortGamesComboBox.addItems(['By Players', 'By Game Quality', 'By avg. Player Score'])
+        self.sortGamesComboBox.currentIndexChanged.connect(self.sortGamesComboChanged)
 
         self.modList.itemDoubleClicked.connect(self.hostGameClicked)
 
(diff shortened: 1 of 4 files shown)


src/games/_gameswidget.py
```diff
@@ -133,6 +133,10 @@ def __init__(self, client, *args, **kwargs):
 
         self.gameList.setItemDelegate(GameItemDelegate(self))
         self.gameList.itemDoubleClicked.connect(self.gameDoubleClicked)
+        self.gameList.sortBy = 0 # Default Sorting is By Players count
+
+        self.sortGamesComboBox.addItems(['By Players', 'By Game Quality', 'By avg. Player Score'])
```
Sheeo: Minor detail here is that option 3 should be "By avg. Player Rating", since that appears to be what the option does.

src/games/gameitem.py
```diff
@@ -634,9 +634,38 @@ def __lt__(self, other):
...
+            try:
+                for player in self.players :
+                    mean = self.client.players[player]["rating_mean"]
+                    dev = self.client.players[player]["rating_deviation"]
+                    selfScore += mean - 3 * dev
+
+                for player in other.players :
+                    mean = self.client.players[player]["rating_mean"]
+                    dev = self.client.players[player]["rating_deviation"]
+                    otherScore += mean - 3 * dev
```
Sheeo: I would prefer making this a property of GameItem, so you would be able to simplify this code to something like `return self.average_rating > other.average_rating` and only define the computation once.

Sheeo:
Hi HaraldWeber, thanks for the contribution!

I'm glad you made this PR to the `master` branch. We'll probably be able to get it shipped with the upcoming client/server update.

Sheeo:
Good to go, nice work :+1:

PR #210, opened 2015-06-05
Added settings to auto download missing maps and mods. 

HaraldWeber:
To bypass annoying message boxes to confirm map and mod downloads, check the settings for auto downloading in the client settings.

### DIFF
--- a/src/_res/client/client.ui
+++ b/src/_res/client/client.ui
@@ -259,6 +259,8 @@
       <string>Settings...</string>
      </property>
      <addaction name="actionSetAutoLogin"/>
+     <addaction name="actionSetAutoMapDownload"/>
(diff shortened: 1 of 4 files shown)


Sheeo:
@HaraldWeber I'm sorry for the late reply on this.

We're about to release 0.10.123, would you get in touch with me at sheeo@sheeo.dk? I'd like to talk about maintaining the client :)

arkitect-dev:
did anything come from this? should this be closed now? @Sheeo

Sheeo:
Was redone because I forgot about this pr with 96b2630ca99c7049aa2861970cc5b759839e3f84.

Will add the automatic downloading of maps part as well. I don't think they need to be two seperate options actually?

PR #212, opened 2015-07-09
wine Mumble hack - initial commit

Wesmania:
Adds code that detects whether FAF was launched uner wine. Uses it to
hack together a method to pass mumble link calls to Linux Mumble.

The code launches a Python script linux-side and loads a mumble_link
stub module that passes calls to it through sockets. It needs an
appropriate mumble_link linux library to run.

I didn't have the means to test the code, but it's syntactically correct
and shouldn't break the windows build.

### DIFF
--- a/src/FAF_wine_mumble_helper
+++ b/src/FAF_wine_mumble_helper
@@ -1,0 +1,102 @@
+#!/usr/bin/env python2
+
+#-------------------------------------------------------------------------------
+# Copyright (c) 2015 Igor Kotrasinski.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the GNU Public License v3.0
+# which accompanies this distribution, and is available at
+# http://www.gnu.org/licenses/gpl.html
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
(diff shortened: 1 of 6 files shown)


Sheeo:
Indeed this is quite the hack. I don't have the means to test it myself currently, and I don't think it's a good idea to merge this before at _least_ that has been done.

src/FAF_wine_mumble_helper
```diff
@@ -0,0 +1,102 @@
...
+mumble_sock = None
+client_sock = None
+prefix = ""
+
+class ConnectionEnd(Exception):
+    pass
+
+def link():
+    local = "127.0.0.1"
+    port = 6113
```
Sheeo: This is a rather unfortunate choice of local port, being so close to the default that the game itself uses.

Wesmania: It's an initial commit; it's probably much better to move it to a separate file anyway. We can change it to whatever we want.

Wesmania:
It's about as clean as it can get, at least until either wine runs mumble properly (which will take a while), or until FAF client becomes fully platform independent (AND I DON'T WANT TO WAIT :/).

I didn't manage to pack FAF into an exec, but will gladly test it if it's possible to pack it.

Sheeo:
It's not too far from being fully platform independent, I would prefer commits work toward this goal rather than hacking around wine usage.

Take a look at https://github.com/Frozen-byte/lobby/network or https://bitbucket.org/fafafaf/modular-client-linux-fix for a client that runs natively in linux.

Wesmania:
I know, that's why I forked it from the latest stable tag rather than the develop branch. It's just a hotfix hack until the forks that do it the right way fix all the bugs. Either way, up to you.

Sheeo:
All right, thanks for taking the time to hack this together -- I hope it's useful for someone (Or yourself). But I don't think we should include it in the master branch, instead going for native compatibility (And running the game through wine) is a better approach. Closing this PR then :+1:

PR #220, opened 2015-08-09
Improving Mod Vault Searching

CodingSquirrel:
We no longer do a local search when typing in the search input box.
We now search when you press enter in the search input box.


ChrisKitching:
:D

ChrisKitching:
... That was probably the right branch. Sheeo? I think develop is "crazy new-patcher land" for now?

Sheeo:
Yes that's fine.

I would prefer we also clean up the localSearch function, which is (presumably?) no longer used.

Bonus points for implementing google-like instant search ;)

CodingSquirrel:
I removed the localSearch function. I'm not quite sure what you mean.

Sheeo:
I skimmed the code a bit too fast and missed the negative diff removing the function, sorry!

PR #238, opened 2015-09-24
Default game name when settings do not exist

madformuse:
When settings are cleared or client used for first time client.login is
not available early enough.


ChrisKitching:
Try again with some commits ;)

madformuse:
It did have some commits but I reset my fork because Sheeo doesn't want me to make it work but wants to solve the underlying state bug :p

ChrisKitching:
Good choice :D

ChrisKitching:
Get cracking! :D

madformuse:
Yes sir :)

PR #247, opened 2015-10-26
Refactor/server packets

IDragonfire:
Outsource some incoming server packets calls ...
First step to split network layer from ui code ...

Discoveries:
* We have a seperate tournament server
* We have a seperate stats server that use the `handle_*` methods of the client

### DIFF
--- a/src/client/server_packets.py
+++ b/src/client/server_packets.py
@@ -1,0 +1,23 @@
+from PyQt4 import QtCore
+
+class ServerPackets(QtCore.QObject):
+
+    # signals for incoming network packages
+    tutorialsInfo = QtCore.pyqtSignal(dict)  # https://github.com/FAForever/server/search?q=tutorials_info
+    modInfo = QtCore.pyqtSignal(dict)  # https://github.com/FAForever/server/search?q=mod_info
+    gameInfo = QtCore.pyqtSignal(dict)  # https://github.com/FAForever/server/search?q=game_info
+    modVaultInfo = QtCore.pyqtSignal(dict)  # https://github.com/FAForever/server/search?q=modvault_info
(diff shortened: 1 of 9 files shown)


Sheeo:
This looks like a step in the right direction, thanks :)

IDragonfire:
If you have suggestion, ideas contact me in slack ...
The next step is to refactore more and more code that we later can change the ui or the server(s) without any problems :D
Next target, Concept for: 
* Client Actions
* Admin Actions
(https://github.com/IDragonfire/lobby/commits/draft/mode_rebased)

src/client/server_packets.py
```diff
@@ -0,0 +1,23 @@
+from PyQt4 import QtCore
+
+class ServerPackets(QtCore.QObject):
```
Sheeo: I'd prefer if we gave this class a bit more responsibility than simply being a dictionary with an aliased `.get(x, None)` function. (This is what `translate` essentially does).

I'd suggest calling it CommandDispatcher and to move client.dispatch in here as well.

Sheeo:
Like so?

IDragonfire:
Yes, but we should then name it correctly...  `PrimaryCommandDispatcher ` because we need something similar for the stats and tournaments commands.

Sheeo:
These are going away, so no, we do not :)

IDragonfire:
Then you can merge it from my view

Wesmania:
The messages / signals code has been significantly refactored since this PR, so I'll take the liberty to delete the corresponding PR branch.

IDragonfire:
I stopped python client development, do it.
Some years ago under zep I implemented a friend list and refactored this for the friend list...

Issue #100, opened 2014-10-27
Name change


Sheeo:
Original bug report by Lextoc at FAForever/fa/issues/96

If a friend (in FAF) changes their name while you are online yourself they'll not appear as a friend until you re-log.

(Should you have too much time on your hands x)

HardlySoftly:
This can be closed now that friendliness isn't done in an insane way :)

(hasn't been a problem for ages)

PR #248, opened 2015-10-27
Draft/model

IDragonfire:
I have no idea what is the best start to move all the model stuff  (entity) out of the ui code (control).
I am also not sure what is the best phytonic solution ...
Classes with getter or setter or a different approach.
Maybe it is possible to share some models between server and lobby.

But the current status is very bad ...
* hard to reuse/share code 
* hard to change implementation (server and ui side)
* Big classes mixed with control, entity and boundary code

### DIFF
--- a/src/client/client_action.py
+++ b/src/client/client_action.py
@@ -1,0 +1,55 @@
+import client
+from fa.replay import replay
+from chat._avatarWidget import avatarWidget
+from PyQt4 import QtCore
+
+class Client_Action(QtCore.QObject):
+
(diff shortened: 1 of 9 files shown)


IDragonfire:
This can also help for problems like this #100

Wesmania:
While the PR branch for this is stale, let's keep it around - at the very least the first commit might be a good starting point once we refactor the code enough to implement some data model.

IDragonfire:
I dropped python client development, feel free to overtake it...

Issue #72, opened 2014-10-03
apparent inconsistency with settings


faf5678:
in src/fa/play:
if settings.value("fa.play.write_game_log", type=bool):
if settings.value("fa.play.stream_live_replay", type=bool):

"write_game_log" and "stream_live_replay" is not referenced anywhere else in the code.

these lines also raise an exception in wine:

Traceback (most recent call last):
  File "C:\fa\lobby\src\client\_clientwindow.py", line 1743, in process
    self.dispatch(json.loads(action))
  File "C:\fa\lobby\src\client\_clientwindow.py", line 1777, in dispatch
    getattr(self, cmd)(message)
  File "C:\fa\lobby\src\client\_clientwindow.py", line 1910, in handle_game_launch
    fa.play(game_info, self.relayServer.serverPort(), arguments)
  File "C:\fa\lobby\src\fa\play.py", line 50, in play
    arguments = build_argument_list(game_info, port, arguments)
  File "C:\fa\lobby\src\fa\play.py", line 28, in build_argument_list
    if settings.value("fa.play.write_game_log", type=bool):
TypeError: unable to convert a QVariant of type 0 to a QMetaType of type 1

PR #251, opened 2015-10-28
Move Settings Code into own class

IDragonfire:
This is as well a first step ...
I started with `_clientwindow.py` ...
settings are loaded everywhere in the codebase ....

Discoveries:
* https://github.com/FAForever/client/pull/250

### DIFF
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -4,6 +4,7 @@ from client.updater import fetchClientUpdate
 from config import Settings
 import fa
 from fa.factions import Factions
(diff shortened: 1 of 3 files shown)


IDragonfire:
Maybe related to #72

HardlySoftly:
Can we do this?

Sheeo:
The persisted property class I added in config probably supercedes this PR. We need to make the rest of the things use it though.

Sheeo:
With the changes to `config.Settings` I think we can close this. Thanks for your work :)

PR #252, opened 2015-11-01
actually calculate avg rating of a game instead of sum

yorick-ne:
if sort games by avg player Rating option was chosen games would actually be sorted by the total sum of player ratings in the lobby - this commit fixes that

### DIFF
--- a/src/games/gameitem.py
+++ b/src/games/gameitem.py
@@ -404,12 +404,12 @@ class GameItem(QtGui.QListWidgetItem):
 
     @property
     def average_rating(self):
-        rating = 0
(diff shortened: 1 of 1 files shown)


Sheeo:
Hi @yorick-ne, thanks for the PR and the attention to this detail :+1:

In the future please keep in mind our contribution guidelines, namely (from README):

    Pull Requests must have py.test unit test coverage.

I implemented the correction with unit test coverage in https://github.com/FAForever/client/commit/4df02d6e25b55cb870045fbe26abced4d92caedb

Issue #208, opened 2015-05-06
Ensure commandline arguments get properly quoted


Sheeo:
Bug report on the forums [here](http://forums.faforever.com/forums/viewtopic.php?f=3&t=9879)

HardlySoftly:
The command line arguments don't seem to be parsed correctly by FA.exe when they include a space.

For normal .fafreplay files this is fine, since it gets moved to a cache directory path without spaces, but in the case of .scfareplay files it gives FA.exe the raw file location to use.  This inevitably causes issues since the filename in .scfareplay contains spaces (that was the naming scheme).

By renaming one of my .scfa replays to not include spaces, I was able to load it successfully.

HardlySoftly:
More generally, even when correctly loaded by FA .scfareplay files immediately desync.

Supporting them would be nice, but I propose that its low priority since the switch to .fafreplay files was made in March 2012.

PR #256, opened 2015-11-04
Fix replay loading

HardlySoftly:
Removes bad method call, wraps up replay file paths.


HardlySoftly:
Fixes https://github.com/FAForever/client/issues/208

Sheeo:
Thanks :)

PR #263, opened 2015-11-15
Better chat

HardlySoftly:
Adds the ability to leave crucial channels.

Adds a button for rejoining all crucial channels in the event that all channels are closed.

### DIFF
--- a/res/chat/chat.ui
+++ b/res/chat/chat.ui
@@ -68,6 +68,11 @@
     </item>
    </layout>
   </widget>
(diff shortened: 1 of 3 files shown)


Sheeo:
This still isn't working, as we discussed in chat :) I sent you the crash logs. Any progress?

HardlySoftly:
I don't recall discussing this or receiving crash logs.

Sheeo:
Since it's currently definitely _crucial_ to stay in aeolus, I think we should revisit this later.

PR #274, opened 2015-12-19
Display Username instead of UserId in the hosting preview

yorick-ne:
Minor Bugfix,
previosly the UserID was shown instead of the name in the preview/setup window for hosting


yorick-ne:
i didnt mean to include the last two commits, i will try to fix my git and then try again :/

PR #277, opened 2015-12-22
Display Username instead of UserID in the Hostingpreview

yorick-ne:
In the Window where you setup a game before hosting, the ID ofa  user would be displayed instead of the username

PS: 2nd try for this pull request, maybe this time git is on my side :p


Sheeo:
Thanks :)

PR #281, opened 2016-01-03
Autojoin custom IRC-channels on startup

yorick-ne:
Enables automatic joining of channels on startup. The channels to join can be picked in the options on top ( options->chat->Set Autojoin channels->tick or untick channels) . Only channels you are in or that  were in the settings, but that you left since, are shown there.

PS: This is my first bigger commit, so if i did smth wrong or didnt do it stylisticaly right  let me know
PPS: The loading of the Settings is kind of a workaround right now, but it seems to run as intended like that

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -224,6 +224,45 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
     def log_event(self, e):
         self.serverLogArea.appendPlainText("[%s: %s->%s]" % (e.eventtype(), e.source(), e.target()) + "\n".join(e.arguments()))
 
+    def updateAutojoinIRCMenu(self):
+        self.client.menuAutojoinIRC.clear()
+        
+        autojoin_channels = Settings.get("chat/autojoin_channels", [], type=str).split(".")
+        autojoin_channels.remove('')
+        
+        for channel in self.channels:
+            if not self.channels[channel].private and not channel == "#aeolus":    
+                action = self.client.menuAutojoinIRC.addAction(channel)
+                action.setCheckable(True)
+                # We set the state of the items before assigning an function to them, since
+                # otherwise the function would get called to early
+                if channel in autojoin_channels:
+                    action.setChecked(True)
+                action.toggled.connect(lambda checked, chan=channel: self.updateAutojoinSettings(chan,checked))
+                
+        #now add channels we autojoin, but that we left in the meantime
+        for channel in autojoin_channels:
+            if channel not in self.channels:
+                action = self.client.menuAutojoinIRC.addAction(channel)
+                action.setCheckable(True)
+                action.setChecked(True)
+                action.toggled.connect(lambda checked, chan=channel: self.updateAutojoinSettings(chan,checked))
+                
+        
+
+    def updateAutojoinSettings(self, channel, checked):
+        logger.info("Setting Autojoin Settings for channel %s to %s" %(channel, "True" if checked else "False"))
+        currentSettings = Settings.get("chat/autojoin_channels", [], type=str).split(".")
+        if checked:
+            currentSettings.append(channel)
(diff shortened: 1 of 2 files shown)


Sheeo:
@yorick-ne hi, sorry for the slow reply on this! The feature is a nice idea.

I find a few bugs with the implementation though.

I get this exception when running it initially without a preconfigured config file:

```2016-01-29 12:02:13,463 ERROR    __main__                       Uncaught exception
  File "src\chat\_chatwidget.py", line 93, in poll
    self.once()
  File "src\chat\irclib.py", line 1131, in once
    self.ircobj.process_once()
  File "src\chat\irclib.py", line 215, in process_once
    self.process_data(i)
  File "src\chat\irclib.py", line 184, in process_data
    c.process_data()
  File "src\chat\irclib.py", line 587, in process_data
    self._handle_event(Event(command, prefix, target, [m]))
  File "src\chat\irclib.py", line 610, in _handle_event
    self.irclibobj._handle_event(self, event)
  File "src\chat\irclib.py", line 326, in _handle_event
    if handler[1](connection, event) == "NO MORE":
  File "src\chat\irclib.py", line 1055, in _dispatcher
    getattr(self, m)(c, e)
  File "src\chat\_chatwidget.py", line 417, in on_privnotice
    self.on_identified()
  File "src\chat\_chatwidget.py", line 283, in on_identified
    self.autoJoin(Settings.get("chat/autojoin_channels", [], type=str).split("."))
AttributeError: 'list' object has no attribute 'split'
```

src/chat/_chatwidget.py
```diff
@@ -224,6 +224,45 @@ def join(self, channel):
     def log_event(self, e):
         self.serverLogArea.appendPlainText("[%s: %s->%s]" % (e.eventtype(), e.source(), e.target()) + "\n".join(e.arguments()))
 
+    def updateAutojoinIRCMenu(self):
+        self.client.menuAutojoinIRC.clear()
+        
+        autojoin_channels = Settings.get("chat/autojoin_channels", [], type=str).split(".")
```
Sheeo: The default type and the type argument given by `type=str` should match. So default should be `""`, otherwise `[].split(".")` will fail.

src/chat/_chatwidget.py
```diff
@@ -241,6 +280,7 @@ def on_identified(self):
         #Perform any pending autojoins (client may have emitted autoJoin signals before we talked to the IRC server)
         self.autoJoin(self.optionalChannels)
         self.autoJoin(self.crucialChannels)
+        self.autoJoin(Settings.get("chat/autojoin_channels", [], type=str).split("."))
```
Sheeo: Ditto (see above)

src/chat/_chatwidget.py
```diff
@@ -224,6 +224,45 @@ def join(self, channel):
...
+                action = self.client.menuAutojoinIRC.addAction(channel)
+                action.setCheckable(True)
+                action.setChecked(True)
+                action.toggled.connect(lambda checked, chan=channel: self.updateAutojoinSettings(chan,checked))
+                
+        
+
+    def updateAutojoinSettings(self, channel, checked):
+        logger.info("Setting Autojoin Settings for channel %s to %s" %(channel, "True" if checked else "False"))
+        currentSettings = Settings.get("chat/autojoin_channels", [], type=str).split(".")
```
Sheeo: Ditto. At this point, since this line is duplicated three times, I suggest you use the `persisted_property` helper I've started using elsewhere in the codebase. Then you can pull out the settings getter into a property, e.g. `ChatWidget.autojoin_channels` which gets 'automatically' persisted when assigned.

If you wrap the persisted property with another property, you could make syntactic sugar to e.g. allow `self.autojoin_channels += ['#my_favourite_room']`

yorick-ne:
I just pushed an update for this in #287 (not here since me and git dont get along that well :) )

PR #287, opened 2016-02-04
Autojoin custom, language and newbie channels on startup

yorick-ne:
Some Stuff based on #281 , due to some git fuckup in another pull request.
- Fixed the Bugs of #281 and added the possibility to save lists of String with persitent Settings.
- Added Language and Newbie channels with respective options in the menu.

The language channels and the deviation threshold for the newbie channel ( anyone know a better name for this channel?)  certainly need some finetuning, but i didnt really know what values is should expect there so i left them in my testing state ( probably everyone will join the newbie channel right now ).

### DIFF
--- a/res/client/client.ui
+++ b/res/client/client.ui
@@ -288,12 +288,12 @@
      <property name="title">
       <string>Chat...</string>
      </property>
-     <widget class="QMenu" name="menuAutojoinIRC">
+     <widget class="QMenu" name="menuAutoJoinIRC">
       <property name="title">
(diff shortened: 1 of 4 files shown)


yorick-ne:
This is now rebased onto some later commit and i added a commit to get rid of that temporary variable when removing a channel. Let me know if anything is missing.

Travia21:
Is there any update to this? I was going to make this my first contribution, but I was informed of this issue/pull request.

PR #288, opened 2016-02-05
Linux support

muellni:
I made an [Arch Linux package for the FAF client](https://aur.archlinux.org/packages/python2-fafclient/) and did some changes to the client to make it work.


Sheeo:
Hey mullni, thanks for the contribution! I won't have time to review this right now but having an arch package is awesome. I'll get back in touch with you as soon as possible.

muellni:
Hi Sheeo, thanks for the comment. Take your time, I know you're pretty busy ATM. 
I really appreciate the work you do for my favorite game. Thank you!
Having the client releases on Github is a huge step forward for me and my friends using Linux.

visionik:
Muellni, you have access to OS X?   I'd pay a small bounty if you can get this working on OS X with wine, too.

muellni:
> Muellni, you have access to OS X?

No, I have not. I think you'd need permanent access for OS X packaging and mine would only be temporarily.
But I think you could find a FAF player who boots Windows for playing.

Playing FA in wine works quite well, but the graphics is not as sophisticated as in Windows.

Sheeo:
@muellni I don't think packaging on the app store would work well.

@visionik work with wine on linux should be pretty much directly transferable to OS X.

muellni:
> @muellni I don't think packaging on the app store would work well.

I was thinking about something like [Homebrew](http://brew.sh/) which seems to provide wine and python packages. But I wouldn't maintain a package I cannot test upon updates.

Sheeo:
I can't really test this until I setup a linux vm, apologies about the delay.

If someone with linux handy would help out that'd be great :+1: (It's not that I do not trust you @muellni, I would just like to make sure)

b2ag:
Hi Sheeo, hi Visionik, I'm in close contact with Martin (since ages) and also use his package on Arch Linux. Just to let you know ;)

muellni:
> Is there a reason this is here?

When I use ```wine_prefix = settings.value("wine/prefix", type=str)``` then the test run runs into an exception:
```
  File "/usr/lib/python2.7/site-packages/_pytest/config.py", line 319, in _importconftest
    mod = conftestpath.pyimport()
  File "/usr/lib/python2.7/site-packages/py/_path/local.py", line 650, in pyimport
    __import__(modname)
  File "/home/sws/projPriv/2016/fafclient/conftest.py", line 10, in <module>
    import config
  File "/home/sws/projPriv/2016/fafclient/src/config/__init__.py", line 81, in <module>
    VERSION = version.get_git_version()
  File "/home/sws/projPriv/2016/fafclient/src/config/version.py", line 82, in get_git_version
    release_version = read_release_version()
  File "/home/sws/projPriv/2016/fafclient/src/config/version.py", line 54, in read_release_version
    import util
  File "/home/sws/projPriv/2016/fafclient/src/util/__init__.py", line 91, in <module>
    "Gas Powered Games", "Supreme Commander Forged Alliance")
  File "/usr/lib/python2.7/posixpath.py", line 70, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'QString' object has no attribute 'endswith'
```
With ```wine_prefix = str(settings.value("wine/prefix", type=str))``` it starts fine. ``` type=str``` seems to mean ```QString``` instead of python string.

muellni:
Maybe it would help you with the review if I restart my changes and create several smaller pull requests instead of this large one?

Sheeo:
I do prefer to review commit by commit, but it's not all that important. I have some more questions and I'll just add them to the big diff.

src/config/version.py
```diff
@@ -51,7 +51,9 @@ def is_prerelease_version(version):
 
 def read_release_version():
     try:
-        f = open("RELEASE-VERSION", "r")
+        import util
```
Sheeo: Because this imports util and we happen to use this file as a script (This itself is probably bad practice), the appveyor build has lost its version number: https://ci.appveyor.com/project/Sheeo/client/build/392#L147

setup.py
```diff
@@ -1,120 +1,146 @@
...
+    )
+else:
+    from distutils.core import setup
+    from setuptools import find_packages
+    product_name = 'fafclient'
+    packages=find_packages()
+    print(packages)
+    setup(
+        name=product_name,
+        version=os.getenv('FAFCLIENT_VERSION'),
```
Sheeo: Where do you set this variable?

Preferably, we should use a build script that sets the version consistently across our build environments (travis ci and appveyor).

muellni: I set this in the [Arch build script](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python2-fafclient) with
```
FAFCLIENT_VERSION=$pkgver python2 setup.py build
```

src/config/version.py
muellni: I decided to install the python files to ```/usr/lib/python2.7/site-packages/``` and the ```res``` folder to ```/usr/share/fafclient```. 
This works pretty well because the ```util.COMMON_DIR``` variable is used everywhere (but not in the config/version.py script). Maybe this needs further refactoring and extraction of the paths from the util script.

setup.py
Sheeo: All right; I think we want to get rid of all of the other version information and simply provide it externally through env during build. This means the build script should be responsible for writing RELEASE-VERSION directly, and the 'live' code should always read this.

How does that sound?

muellni: > How does that sound?

Fine I guess.
My build scripts writes RELEASE-VERSION.

Crotalus:
:bell:

muellni:
I think we can close this PR now. I'm currently using [my linux_support_2 branch](/muellni/client/tree/linux_support_2).

PR #293, opened 2016-02-08
Fix private messages from Foes

yorick-ne:
Previously Foes could send private messages  ( and "private actions").
Also Foes blend in with the backgorund, but their messages can be seen when selecting their text.

foes can however still ping one by typying ones name in a chatroom, this will need a fix at  another time


yorick-ne:
Related to this: https://github.com/FAForever/client/blob/develop/src/chat/channel.py#L286
I dont think this will ever fail, i.e. foes will still be able to write visible messages.
The Issue should be the same as above, isFoe expects an id of a player not its name

yorick-ne:
yeah, i think i approached this from the wrong direction
i had another look at the isFoe and isFriend functions earlier and it appears to me like they shoulndt store the id, but instead the name / or the functions should take the name as the parameter atleast
i was basicaly adapting to a bug the whole time with this pull request xD

PR #294, opened 2016-02-08
Fixed connectivity is_ready function

Blackclaws:
QAbstractState is actually a function state() return the state.


Sheeo:
Nice job tracking this down.

PR #297, opened 2016-02-11
Call isFoe more consistently with id and rename stuff inside isFriend…

yorick-ne:
… and isFoe methods
This should fix foeing in irc, i.e. preventing pm's and not showing what they write in chat channels.
Some stuff based on the game_info message is still not working, for example sorting games and notifications when hosting


Sheeo:
Nice work, thanks :)

PR #298, opened 2016-02-12
Process 'Rehost'

micheljung:
Let the client process 'Rehost' messages and request a rehost from the server.

### DIFF
--- a/src/fa/game_session.py
+++ b/src/fa/game_session.py
@@ -29,6 +29,13 @@ class GameSession(QObject):
     def __init__(self, client, connectivity):
         QObject.__init__(self)
         self._state = GameSessionState.OFF
+        self._rehost = False
+        self.game_uid = None
+        self.game_name = None
+        self.game_mod = None
+        self.game_visibility = None
+        self.game_map = None
+        self.game_password = None
 
         # Subscribe to messages targeted at 'game' from the server
         client.subscribe_to('game', self)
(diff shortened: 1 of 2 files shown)


src/fa/game_session.py
```diff
@@ -147,6 +154,9 @@ def _on_game_message(self, command, args):
             elif args[0] == 'Lobby':
                 # TODO: Eagerly initialize the game by hosting/joining early
                 pass
+        elif command == 'Rehost':
+            self._rehost = True
```
Sheeo: This appears to be a write only property?

src/fa/game_session.py
```diff
@@ -161,3 +171,19 @@ def _exited(self, status):
...
+        if self._rehost:
+            self.client.host_game(title=self.game_name,
+                                  mod=self.game_mod,
+                                  visibility=self.game_visibility,
+                                  mapname=self.game_map,
+                                  password=self.game_password,
+                                  is_rehost=True)
+
+        self._rehost = False
+        self.game_uid = None
```
Sheeo: How about we store these in a dict instead?

Sheeo:
I'll make the suggested amendments when merging this, sorry for the delay

src/fa/game_session.py
micheljung: No, see line 175; `if self._rehost`

src/fa/game_session.py
micheljung: I guess a dict would be ok, I'm just not a fan of losing the last bit of of compile time checks and auto completion one can get with python x) (it's easy to misspell a dict-key and you have to manually check how the one you need to set is named)

micheljung:
Sure, no problem. I know you have other priorities busy atm.

micheljung:
Related to https://github.com/FAForever/fa/pull/1330

micheljung:
This shouldn't have been merged; the suggested amendments were never made, it was never tested and it doesn't even work.

PR #300, opened 2016-02-20
Add an exploit warning to the games tab

Blackclaws:
I think this is a nice addition as neither the chat nor the news page contains a link to the exploit forum thread anymore and I've been hearing reports of people simply not knowing that there is a list of exploits (noone checks the forum if they just use the client) especially if they have been away for a while.

I think we might still tweak the way it is styled right now and also the link isn't working for some reason. But I wanted to get a first shot out.

### DIFF
--- a/res/games/games.ui
+++ b/res/games/games.ui
@@ -602,6 +602,95 @@ p, li { white-space: pre-wrap; }
        </widget>
       </item>
       <item>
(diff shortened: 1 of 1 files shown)


Sheeo:
With 3652 hopefully being released soon this is somewhat obsolete unfortunately

PR #301, opened 2016-02-21
Rename Find Games tab to Play

speed2CZ:


HardlySoftly:
What's the reason for this?

speed2CZ:
people that ask where can they host games

HardlySoftly:
Right, although I don't necessarily think this will help.  Also it sort of implies that Coop tab doesn't mean the same thing as playing...

speed2CZ:
Let's put it this way, I don't think it can make things worse. Also many people were suggesting this.

HardlySoftly:
I think it would be best to merge the coop tab into the "Play" tab, and have one place for all game related things.

speed2CZ:
And move coop leaderboards into leaderboard tab

PR #303, opened 2016-02-21
Changed to user agent used when accessing the map vault to FAForever

Blackclaws:
This is needed so the server recognizes the client as the lobby.

Its actually not a big problem in the mainline client but it might cause bugs and is needed for runs from source.


Sheeo:
:+1:

PR #308, opened 2016-02-25
Check on ReplayWidget if the correct item is pressed

TimWolters:
If a user right-clicks in the replayvault on a bucket (with the date of the replays) an error is thrown. This fixes this error with an extra check if an replay is clicked or the bucket.


Sheeo:
Thanks :+1:

Sheeo:
Protip: You shouldn't use `develop` for pull requests, it will create troubles for yourself when the mainline develop branch goes 'out of sync' with your own.

TimWolters:
Thank you, I will remember to use a different branch in the future

PR #312, opened 2016-03-01
Sort the userlist with focus on usefulness

Crotalus:
99% of the time when taking a glance at the userlist you want to see if a
certain friend is online / playing / search in vault etc. The white noise is just a distraction.

If you for some reason have to communicate with a moderator they are easy
to find by scrolling down.  New users will see the moderators due to not having any friends anyway.

This will also tone down the "moderators vs aeolus" feeling when you don't have them watching over you all the time.

### DIFF
--- a/res/client/client.ui
+++ b/res/client/client.ui
@@ -291,6 +291,7 @@
       <string>Chat...</string>
      </property>
      <addaction name="actionColoredNicknames"/>
+     <addaction name="actionFriendsOnTop"/>
      <addaction name="actionSetJoinsParts"/>
      <addaction name="actionSetOpenGames"/>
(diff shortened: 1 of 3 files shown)


Sheeo:
If we do this it should be an option. People often ask who moderators are despite this being obvious from them being on top in white -- I believe hiding them will make this worse.

Please let me know if you intend to progress with this or I should close the PR

IDragonfire:
@crotalus Any Updates?

Crotalus:
Added an option "Friends on top" to Options / Chat

HardlySoftly:
Right now mods that are also friends are placed with mods, and not with friends.

If the point is to move friends to the top, then shouldn't mod friends be moved up with regular friends?

Issue #284, opened 2016-01-20
Suggestion: Select a subset of the factions when searching for  ladder game


yorick-ne:
Currently the Ladder Buttons behave like radiobuttons, i'd suggest to change the behavior analog to the one of checkboxes. With this you could search as UEF and Cybran at once for example.
This would proably need server side support.

Sheeo:
Sounds a lot better than a "Random but not UEF" button, which I've heard a bit about :+1: :)

arkitect-dev:
closing as duplicate

PR #317, opened 2016-03-16
Selecting a subset of factions when searching for a 1 vs 1 match

TimWolters:
Rather straight forward feature. The user has the option to select in the option menu to change the behaviour of the buttons by being able to select multiple factions and then press play. The client will choose a faction for the user and start searching. 

Existing code has been changed by this as startRankedSearch does not check/uncheck the buttons anymore but is moved to toggle_search.
For creating the ui another method has been added that takes care of the setting the correct handlers on the buttons when the option in the option menu has been selected. It's also called from the constructor.
All changes are saved in the settings file of the user (if he uses the subset selection and the factions selected).

Sadly i've been unable to get the tests to run for a while now so it's only been tested by hand.

### DIFF
--- a/src/games/_gameswidget.py
+++ b/src/games/_gameswidget.py
@@ -151,6 +151,7 @@ class GamesWidget(FormClass, BaseClass):
         if(self.use_subset):
             self.rankedPlay.clicked.connect(self.startSubRandomRankedSearch)
             self.rankedPlay.show()
+            self.rankedRandom.hide()
             self.labelRankedHint.setText("Select a few of your favorite factions and hit play to get matched against a player of your level in a random map !")
             for faction, icon in self._ranked_icons.items():
                 if(disconnect):
@@ -160,6 +161,7 @@ class GamesWidget(FormClass, BaseClass):
                 icon.clicked.connect(self.selectSubset)
         else:
             self.rankedPlay.hide()
+            self.rankedRandom.show()
             self.labelRankedHint.setText("Select your favorite faction and get matched against a player of your level in a random map !")
             for faction, icon in self._ranked_icons.items():
                 if(disconnect):
--- a/src/games/_gameswidget.py
+++ b/src/games/_gameswidget.py
@@ -15,13 +15,19 @@ import notifications as ns
 from config import Settings
 
 import logging
+
 logger = logging.getLogger(__name__)
 
 FormClass, BaseClass = util.loadUiType("games/games.ui")
 
+PLAY_SUBSET_FACTIONS = "play/subsetFactions"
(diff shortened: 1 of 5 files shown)


TimWolters:
Implementation for #284

src/games/_gameswidget.py
```diff
@@ -12,13 +12,23 @@
...
 FormClass, BaseClass = util.loadUiType("games/games.ui")
 
+PLAY_SUBSET_FACTIONS = "play/subsetFactions"
+PLAY_SUBSET_SELECT = "play/selectSubset"
+
+
 class GamesWidget(FormClass, BaseClass):
+
+    Settings.persisted_property(PLAY_SUBSET_SELECT, type=bool, default_value=False)
+    Settings.persisted_property(PLAY_SUBSET_FACTIONS, default_value=[False, False, False, False, False])
```
Sheeo: Ditto on this (As here: https://github.com/FAForever/client/pull/318#discussion-diff-57013181), you'll want to assign these properties for intended use.

Sheeo:
So I ended up playing a ladder match when testing this out, woops.

I couldn't immediately figure out how subset searching works though -- I think we should make it more obvious in the UI instead of hiding it behind an option. What do you think?

Also, when searching for a subset -- we should consider excluding 'random', since that is sort of confusing?

Nice work :+1:

TimWolters:
Thank you for taking the time with all these reviews.

I share the idea that the ui is not clear. I already used some pointers from Exotic_Retard [(here)](http://forums.faforever.com/viewtopic.php?f=45&t=11969). Still there has to be a way to make the start button more clickable (if that makes sense). Also the option in the menu doesn't make it clear what it modifies. I could make it a show/hide button next to random as I did previously.
It makes sense to hide the random button while using subset and I'm doing that next.

We could also leave the start button out of it and let the player add and remove factions(as long as one is selected) while searching but that would modify existing code even more and I initially wanted to avoid that.

TimWolters:
Even though I pushed this I'm not really happy with it. First of all I have to manually set the settings and the first element of sub_factions will never be changed in the settings file. I've been looking for a while for a reason on both of the problems but I haven't found it (yet).

IDragonfire:
@TimWolters Any Updates?

TimWolters:
I have some problems with saving the selected options to the settings file. Work on this was already done. So I was waiting on that being merged and than this is done. I've no ideas on the UI to make it more on making it more obvious as suggested by Sheeo.

Crotalus:
Merged this to `develop` without making it optional. Select all / none factions to make it use random faction when searching ladder.

PR #318, opened 2016-03-16
Save settings of sorting games and hiding private games in the user configuration file

TimWolters:
Saves the state of the sort games and hide private games in the play tab.
This is currently not saved in the FA Lobby.ini and I think it would be nice if it did.

### DIFF
--- a/src/games/_gameswidget.py
+++ b/src/games/_gameswidget.py
@@ -16,17 +16,15 @@ import notifications as ns
 
 import logging
 
-SORT_GAMES = "play/sortGames"
-HIDE_PRIVATE_GAMES = "play/hidePrivateGames"
-
 logger = logging.getLogger(__name__)
 
 FormClass, BaseClass = util.loadUiType("games/games.ui")
 
-Settings.persisted_property(HIDE_PRIVATE_GAMES, default_value=False, type=bool)
-Settings.persisted_property(SORT_GAMES, default_value=0, type=int)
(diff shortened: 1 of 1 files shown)


Sheeo:
Hey, sorry for the late review, I'll comment in the code

src/games/_gameswidget.py
```diff
@@ -14,10 +15,17 @@
...
+
+SORT_GAMES = "play/sortGames"
+HIDE_PRIVATE_GAMES = "play/hidePrivateGames"
+
 logger = logging.getLogger(__name__)
 
 FormClass, BaseClass = util.loadUiType("games/games.ui")
 
+Settings.persisted_property(HIDE_PRIVATE_GAMES, default_value=False, type=bool)
+Settings.persisted_property(SORT_GAMES, default_value=0, type=int)
```
Sheeo: This is meant to be used as a property on a class, calling these functions like this initializes a setter/getter but then immediately discards it -- that doesn't seem right.

You'll want to assign it on the GamesWidget class like so:

```
class GamesWidget(FormClass, BaseClass):
    ...
    sort_games = Settings.persisted_property(...)
```

After which `self.sort_games = index` will automatically persist it, and `self.sort_games` will yield the value from configuration.

Sheeo:
Merged with f0f5f56121914527dc1acd75c4298ad77cb25670

Issue #309, opened 2016-02-25
pressing abort if port are occipied doesnt exit faf


yorick-ne:
if there already happens to be an instance of faf running and the neccesary ports cant be opened faf will ask you in a popup window if you want to continue, faf will start anyway even if you press abort

Wesmania:
Still reproducible in 0.16.0-rc.2.

PR #319, opened 2016-03-22
Exit when port is in use

TimWolters:
Client now does something when doConnect fails as previously it would continue starting the client even though there was no chance on recovery.

Fixing #309


TimWolters:
Fixing #309

Sheeo:
:+1:

PR #329, opened 2016-04-09
Fix Announcements when friends host/play a game

yorick-ne:
This are some clientside adaptation to https://github.com/FAForever/server/pull/119. The only other time the changed gameinfo message would would have an impact (as far as i could see) was in the notification system, which should now work as intended.
I wasnt able to run the server code, so a second look if i missed smth would be appreciated.

### DIFF
--- a/src/games/gameitem.py
+++ b/src/games/gameitem.py
@@ -7,6 +7,7 @@ import os
 from games.moditem import mod_invisible, mods
 
 import client
+from client import Player
 import copy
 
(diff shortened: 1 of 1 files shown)


Sheeo:
Still unclear if the server side change will have more effects than this. Rescheduling for 0.11.60

Crotalus:
:bell: Should try to get announcements back up again, is a vital part of getting games going.

Crotalus:
Fixed in ee925aac10ce2511e89f3e3795eb8a0aa7dae3d2

PR #330, opened 2016-04-09
Pass an ID over when opening a private chat

### LINKED ISSUE #291: Green chat highlight not consistent
It says quit.
It gives you a ping like you got pm'd, but doesn't tell you where. If you have 20 tabs open like I do a lot, that's really annoying XD 
Also some interesting stuff with green:

testalsogreen
greentext
testalsogree2

yorick-ne:
Due to this chatter get their respective color in private chats instead of the default irc dark grey. However the text can be white/blue/green/etc depending on friend/etc status, dunno if that is the intended state or just plain light gray for everything in private messages.
Depending on Interpretation this fixes #291 , atleast it should make it consistent :P

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -183,7 +183,7 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
                     self.channels[channel].printAction("IRC", "was disconnected.")
             return False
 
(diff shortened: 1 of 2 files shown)


Crotalus:
1abb986167e81fd18c4dddd70713d2ae02be84a1

PR #331, opened 2016-04-13
Show what map people are playing / lobbying in nickname list

Crotalus:
I like miniature Seton's icon spam instead of crossed swords spam.

Should be seen as a work in progress at this point,  need some way to show if a player is in a lobby or in an actual game. Maybe put the current used sword icon next to the map icon? 

Open for suggestions :bulb: 
faf

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -10,6 +10,8 @@ from chat import user2name
 from config import Settings
 
 from fa.replay import replay
+from fa import maps
+
 import util
 
 import client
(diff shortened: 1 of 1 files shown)


IDragonfire:
From my view make it configurable in the options ;)

Crotalus:
Yeah will add it as an option after my eyes have healed after the latest venture in the client options code :)

ChrisKitching:
Sword on top of map seems like it'd work. You might need to make a thinner sword icon for this to work well.

After all, you're not expecting to show meaningful detail of the map, just an indication of which map it is. The tiny icon you've got here definitely isn't going to shed any meaningful light on a map you're not familiar with, but it'll let you see which map it is if it is one you recognise. As a result, I don't think it'll do any harm to partially cover it.

duk3luk3:
Do we still want to do this? @Crotalus

Crotalus:
Need to come up with some clever way of showing lobby vs in game if we want these icons

duk3luk3:
display icon with less saturation / contrast / opacity while in lobby?

src/chat/chatter.py
```diff
@@ -183,15 +185,25 @@ def update(self):
...
                 self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
-        else:
-            self.statusItem.setIcon(QtGui.QIcon())
-            self.statusItem.setToolTip("Idle")
+
+            if icon_url:
+                mapname = url.queryItemValue('map')
+                icon = maps.preview(mapname)
+                if not icon:
+                    self.lobby.client.downloader.downloadMap(mapname, self)
```
aeoncleanse, line 200: Won't this download all maps being played by anyone online at a given time?

ChrisKitching: Yup.

For this feature to work sanely, you're going to want some way to get map icons without getting the map itself. The new API probably does that?

aeoncleanse: Not a clue. First thing that jumps to mind as a sane, if not comprehensive or complete, way to do it would be to include a map icons download package with the client installer by default, including most of the common ones, and just use the swords if you don't have them? 

API route to grab by mapname would be better of course.

ChrisKitching: > API route to grab by mapname would be better of course.

Really, this is the only sane approach. An API route to serve up the icon. For bonus points, you could request it at a certain size and have the server cache requested images.

>  include a map icons download package with the client installer by default, including most of the common ones, and just use the swords if you don't have them?

You could. But adding an API route - if it doesn't already exist - should be very straightforward. I'd suggest going right for the final solution.

Wesmania: AFAIK the "downloadMap" method actually downloads the map preview, confusing as it is. The preview images are simply served from content.faforever.com. If you want to add some fancy API routes, then go ahead :)

aeoncleanse: Having something named "downloadMap" only grab the preview image is _fucking stupid_. If true, that needs renaming pronto.

PR #332, opened 2016-04-13
Add option to have Notifications pop up in each corner

yorick-ne:
Adds options for each corner in the Settings for the Notifications stuff, they are also being saved.

### DIFF
--- a/src/notifications/ns_dialog.py
+++ b/src/notifications/ns_dialog.py
@@ -6,7 +6,7 @@ The UI popup of the notification system
 """
 FormClass, BaseClass = util.loadUiType("notification_system/dialog.ui")
 class NotificationDialog(FormClass, BaseClass):
-    def __init__(self, client, *args, **kwargs):
+    def __init__(self, client, settings, *args, **kwargs):
         BaseClass.__init__(self, *args, **kwargs)
 
         self.setupUi(self)
@@ -15,12 +15,8 @@ class NotificationDialog(FormClass, BaseClass):
         self.labelIcon.setPixmap(util.icon("client/tray_icon.png", pix=True).scaled(32, 32))
         self.standardIcon = util.icon("client/comment.png", pix=True)
 
-        screen = QtGui.QDesktopWidget().screenGeometry()
-        dialog_size = self.geometry()
-
-        # TODO: more positions
-        # bottom right
-        self.move(screen.width() - dialog_size.width(), screen.height() - dialog_size.height())
+        self.settings = settings
+        self.updatePosition()
 
         # Frameless, always on top, steal no focus & no entry at the taskbar
         self.setWindowFlags(QtCore.Qt.ToolTip)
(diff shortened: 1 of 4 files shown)


Sheeo:
Appveyor seems to have barfed initially on this one. Scheduled a rebuild.

ChrisKitching:
Appveyor has ceased to barf.
Coveralls is angry because _gasp_ you added more code.

I think this is good to merge?

src/notifications/ns_dialog.py
```diff
@@ -57,3 +53,17 @@ def hide(self):
     def mousePressEvent(self, event):
         if event.button() == QtCore.Qt.RightButton:
             self.hide()
+
+    def updatePosition(self):
+        screen = QtGui.QDesktopWidget().screenGeometry()
+        dialog_size = self.geometry()
+        position = self.settings.popup_position  # self.client.notificationSystem.settings.popup_position
```
ChrisKitching, line 60: What's this comment for?

ChrisKitching, line 60: Oh, right, to express this in ZeP-insanity-parent-pointer-language. Blarg. More restructuring needed. :P

src/notifications/ns_settings.py
```diff
@@ -43,21 +43,26 @@ def __init__(self, client):
     def loadSettings(self):
         self.enabled = Settings.get('notifications/enabled', True, type=bool)
         self.popup_lifetime = Settings.get('notifications/popup_lifetime', 5, type=int)
+        self.popup_position = Settings.get('notifications/popup_position', "bottom right", type=str)
 
         self.nsEnabled.setChecked(self.enabled)
         self.nsPopLifetime.setValue(self.popup_lifetime)
+        comboboxIndex = [i for i in range(4) if self.nsPositionComboBox.itemText(i) == self.popup_position]
+        self.nsPositionComboBox.setCurrentIndex(comboboxIndex[0] if comboboxIndex else -1)
```
ChrisKitching, line 51: But 0 is falsey, so if the combobox index is 0 this won't work as expected (and in any case this list comprehension is a slightly strange way of doing this).

ChrisKitching, line 51: I think the key mistake here is a failure to separate combo box labels (a view thing) from the selected position (a model thing). Read up on "Movel/view pattern" if you want to get really detailed, I guess.

Don't store the labels, store some canonical representation. I'd suggest an enum to represent positions, and a map to map enum values to labels, like this:

```
class NotificationPosition(Enum):
    TOP_LEFT = 1
    BOTTOM_LEFT = 2
    BOTTOM_RIGHT = 3
    TOP_RIGHT = 4

    def getLabel(self):
        if self == NotificationPosition.TOP_LEFT:
            return "top left"
        elif self == NotificationPosition.BOTTOM_LEFT:
            return "bottom left"
        elif self == NotificationPosition.TOP_RIGHT:
            return "top right"
        elif self == NotificationPosition.TOP_RIGHT:
            return "top right"
```

For one thing, this makes localisation actually _possible_ without breaking the stored values (you'd just modify this to call into a localisation function. Your way, if they change language (or we relabel a button) the stored setting becomes invalid). It also means you don't need to do this weird list comprehension thing. You can now, given a `NotificationPosition`, just call `getLabel()` on it to get the needed label. You can trivially make a list of labels as a list comprehension, and you avoid storing localised strings in the settings file.

Crotalus:
68c7f89433b72099d36bc7248cafd66e61a6f48b
aa45c7f95b960a7b010bc6b3d2b13ffb144798c4

PR #334, opened 2016-04-28
Save gamelogs as default

Crotalus:
Any particular reason why saving game logs defaults to off? Saving them as default will save us the trouble of informing people to enable it for debugging purposes.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,7 @@
 * Add linux support (thanks @muellni)
 * Add new uid implementation (thanks @muellni)
 * Leaderboard display performance improvement (#438, thanks @grotheFAF)
+* Save game logs as default
 
 0.11.60
 =======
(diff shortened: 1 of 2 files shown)


Sheeo:
Performance reasons -- from memory the game performs slightly worse because of hitting the disk when logging is on. Especially in the case of logspam in larger games?

Sheeo:
Considering that we quite often lack game logs; it might be worth the trade off. Comments?

Crotalus:
Yeah it would probably be worth it, the log spam in games are a lot less now overall (unless something went awfully wrong).

aeoncleanse:
Yes. That is all. Even better, can we save a rolling cycle of, say, the past 10 games played?

Sheeo:
Sure. You up for doing this @aeoncleanse?

aeoncleanse:
Is Python right? I can have a looj with some handholding probably

Crotalus:
Maybe zip and submit log to server after game if an non-ui error is detected. Would have to remove all current known sim errors first though (rotators, animations etc)

duk3luk3:
We should get this in, shouldn't we? Care to rebase, @Crotalus? otherwise I'll go try the scary red merge button....

Crotalus:
Rebased

duk3luk3:
Oh, and add a changelog entry.

PR #336, opened 2016-05-14
Use the new uid implementation

muellni:
This PR requires https://github.com/FAForever/server/pull/121 on the server!
The client now needs the prebuilt binary Python module from https://github.com/muellni/uid/releases .


Sheeo:
This appears to fail on both CI's -- the `uid` package isn't being installed on any of them.

muellni:
> This appears to fail on both CI's -- the uid package isn't being installed on any of them.

The `uid` module is currently not part of the client. I've uploaded prebuilt binaries to https://github.com/muellni/uid/releases

Sheeo:
To make CI not break you'll need to get it to download it during the build process, then.

ChrisKitching:
This has been merged successfully into the `pyqt5` branch. It may or may not be actually merged into develop on its own before pyqt5 makes its way there.

Thanks for the excellent work. Very glad to be rid of that awful binary, and to be able to do a proper 64 bit build.

ChrisKitching:
Since this has merged (although not quite yet into develop) I guess we can close this to get it out of the way.

Thanks again for saving us from that awful binary :D

PR #338, opened 2016-05-14
Added Window's version of the export command

Travia21:
Just trying to help new people who want to get into dev on Windows


RELEASE-VERSION
```diff
@@ -0,0 +1 @@
+0.11.58+464
```
Sheeo, line 1: This file is generated by the build process, so we shouldn't have it in the source repository here.

I'm aware there's a problem when this file is lacking though; we need that particular issue resolved.

readme.md
```diff
@@ -32,7 +32,8 @@ Running
 -------
 Before running unit tests, set PYTEST_QT_API as follows:
 
-    export PYTEST_QT_API='pyqt4v2'
+    *nix: export PYTEST_QT_API='pyqt4v2'
+	Windows: SET PYTEST_QT_API='pyqt4v2'
```
Sheeo, line 36: Even on windows we recommend using a bash shell, since that's a lot simpler than using cmd.exe along with its associated quirks.

RELEASE-VERSION
Travia21, line 1: It turns out that wasn't the issue. The wiki said to change a line about the version to a hard number caused the window to pop-up.

readme.md
Travia21, line 36: Sweet. I'll add that to the wiki.

PR #339, opened 2016-05-14
Refactor HostgameWidget

yorick-ne:
Moved lots of similar code from GamesWidget and CoopWidget over to HostgameWidget
This should fix hosting for coop games
Also fixes a bug that prevented joining coop games ( untested)

### DIFF
--- a/src/games/hostgamewidget.py
+++ b/src/games/hostgamewidget.py
@@ -108,10 +108,6 @@ class HostgameWidget(FormClass, BaseClass):
             self.password = None
         self.save_last_hosted_settings()
 
-
-        self.save_last_hosted_settings()
-
-
         # Make sure the binaries are all up to date, and abort if the update fails or is cancelled.
         if not fa.check.game(self.parent.client):
             return
--- a/src/games/hostgamewidget.py
+++ b/src/games/hostgamewidget.py
@@ -94,12 +94,14 @@ class HostgameWidget(FormClass, BaseClass):
     def updateText(self, text):
         self.message['title'] = text
         self.game.update(self.message, self.parent.client)
+        self.game.setHidden(False)
 
     def hosting(self):
         name = self.titleEdit.text().strip()
         if len(name) == 0:
             # TODO: Feedback to the UI that the name must not be blank.
             return
+        self.title = name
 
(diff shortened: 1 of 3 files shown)


src/games/hostgamewidget.py
```diff
@@ -86,19 +101,56 @@ def hosting(self):
...
         if self.passCheck.isChecked():
-            password = self.passEdit.text()
+            self.password = self.passEdit.text()
         else:
-            password = None
+            self.password = None
+        self.save_last_hosted_settings()
+
+
+        self.save_last_hosted_settings()
```
Sheeo: This seems redundant?

src/games/hostgamewidget.py
```diff
@@ -86,19 +101,56 @@ def hosting(self):
...
 
     def mapChanged(self, index):
-        self.parent.gamemap = self.mapList.itemData(index)
+        self.mapname = self.mapList.itemData(index)
 
-        self.message['mapname'] = self.parent.gamemap
+        self.message['mapname'] = self.mapname
         self.game.update(self.message, self.parent.client)
+
+    def save_last_hosted_settings(self):
```
Sheeo, line 144: Looks like this method still exists on `GamesWidget`, is that intended?

Would like to see these migrated to use the persisted properties sometime in the future, out of scope of this PR though :)

Sheeo:
Looks good, would like to test that it fixes coop hosting on the test server.

src/games/hostgamewidget.py
yorick-ne, line 144: I think i just forgot to remove that method + 'load_last_hosted_settings' on the Gameswidget class. They didnt appear to get called anywhere.

speed2CZ:
When hosting game I cant change title of the lobby, no matter what I type there it uses the one I had before this fix.

HardlySoftly:
I also had the same issue as @speed2CZ

yorick-ne:
@speed2CZ @HardlySoftly fixed now, thanks for the report

yorick-ne:
@arkitect-dev why has this been closed ?

arkitect-dev:
sorry, waffle closed automatically when i dragged it into another column

Sheeo:
👍

PR #344, opened 2016-05-18
client.urls -> client.instance.urls

Crotalus:
I guess this could be the reason for "sword icons" lingering in user list even if games have ended?


Crotalus:
@sheeo check this out if you have time, client.urls should be client.instance.urls like it's used here:

https://github.com/FAForever/client/blob/develop/src/games/gameitem.py#L151

Issue #341, opened 2016-05-15
Ensure IRC userlist is populated with FAF information


Sheeo:
Currently if the client manages to connect to IRC before it has authenticated with the server, it doesn't have the list of online users available when the #aeolus userlist is sent, which means everyone shows up as IRC users.

Either we should update IRC user information when receiving `player_info` messages, or (simply) defer connecting to IRC until we're authenticated with the FAF server (I believe this was done in the past?) 

**Acceptance Criteria**
- Users in the chatlist should no longer wrongfully display as IRC users in any case.

HardlySoftly:
If the IRC user information was received first, chatters were initialised with an id of -1, and so weren't correctly updated.

PR #346, opened 2016-05-21
Make sure chatters get the correct id when updating

HardlySoftly:
Solves issue https://github.com/FAForever/client/issues/341


Sheeo:
I expect this to be rather slow, but if it solves the problem we'll leave performance to later updates I suppose. Thanks 👍

HardlySoftly:
Adding in a name -> id map should be pretty simple, seems like the best way to go here.

PR #355, opened 2016-06-04
Add login -> player map

HardlySoftly:
Also remove extraneous players.login variable

### DIFF
--- a/src/client/players.py
+++ b/src/client/players.py
@@ -18,17 +18,19 @@ class Players:
     """
     def __init__(self, me):
         self.me = me
-        self._players = {}
-        self._warned = False
-        self.login = None
         self.coloredNicknames = False
-        
+
+        # UID -> Player map
+        self._players = {}
+        # Login -> Player map
+        self._logins = {}
+
         # ids of the client's friends
         self.friends = set()
-        
+
         # ids of the client's foes
         self.foes = set()
-        
+
         # names of the client's clanmates
         self.clanlist = set()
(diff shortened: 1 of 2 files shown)


src/client/players.py
```diff
@@ -18,17 +18,19 @@ class Players:
...
         # ids of the client's friends
         self.friends = set()
-        
+
         # ids of the client's foes
         self.foes = set()
-        
+
         # names of the client's clanmates
         self.clanlist = set()
```
ChrisKitching, line 35: Unrelated to this patch, but why isn't this a set of IDs?

src/client/players.py
```diff
@@ -108,15 +110,13 @@ def __getitem__(self, item):
         if isinstance(item, Player):
             return item
         try:
+            # Lets hope that nobody has an integer valued name
```
ChrisKitching, line 113: That is not forbidden by the registration system.

That said,

```
mysql> SELECT id, login FROM login WHERE login REGEXP '^[0-9]+$';
Empty set (0.06 sec)
```

It might be cleaner to have an explicit `by_name(name)` function to do name lookups, and let the operator be ID-only.

HardlySoftly:
I might add a function for doing lookups by name later, but the implicit assumption that names couldn't be int types was already there so I thought I'd flag it up

Sheeo:
Oh. I closed the later pr with the assumption that this hadn't been merged.

PR #357, opened 2016-06-05
Autojoin custom, language and newbie channels on startup

yorick-ne:
Adds option to automatically join language, newbie or custom defined channels. Language channels only work for a predefined set of countries.

rebased #287 with minor modifications

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -28,6 +28,12 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
     irc_host = Settings.persisted_property('chat/host', type=str, default_value='irc.faforever.com')
     irc_tls = Settings.persisted_property('chat/tls', type=bool, default_value=False)
 
+    autoJoinChannels = Settings.persisted_property('chat/autoJoinChannels', type=list, default_value=[])
+    # All countries that have an associated language chat, the 2-letter key will be shown when hovering
+    # over ones Flag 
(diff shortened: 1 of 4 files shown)


src/chat/_chatwidget.py
```diff
@@ -238,6 +273,14 @@ def on_identified(self):
         #Perform any pending autojoins (client may have emitted autoJoin signals before we talked to the IRC server)
         self.autoJoin(self.optionalChannels)
         self.autoJoin(self.crucialChannels)
+        self.autoJoin(self.autoJoinChannels)
+        
+        #this might be timing dependant
+        if self.client.useLanguageChannel and self.client.me.country and self.client.me.country in self.languageChannels:
+            self.autoJoin(self.languageChannels[self.client.me.country])
+
+        if self.client.useNewbiesChannel and 100 < self.client.me.global_rating[1] < 500:
```
yorick-ne, line 282: This currently moves people in the newbie channel if their deviation is > 100, i'm not quite sure how good this value is

Crotalus:
Needs feedback, personally I think a newbie channel could be nice. Maybe restrict it on server side also though, otherwise the trolls will wreak havoc there. I guess smurfs also so probably wait until the smurf-stamper is re-enabled.

Localised channels, not so sure about that. Maybe for the more obscure people who doesn't speak english well (chinese etc). Otherwise we just risk splitting up the community more.

Wesmania:
Obsoleted by chat rewrite.

Issue #352, opened 2016-06-04
Livereplays arent colored according to friend status


yorick-ne:
This https://github.com/FAForever/client/blob/develop/src/replays/_replayswidget.py#L426 doesnt work as 'isFriend' expects an integer but is given a string.

Wesmania:
Can't reproduce in 0.16.0-rc.2.

PR #358, opened 2016-06-05
Improve support for player checks by name

HardlySoftly:
Also remove casting of name to int

### DIFF
--- a/src/client/players.py
+++ b/src/client/players.py
@@ -45,19 +45,25 @@ class Players:
         '''
         Convenience function for other modules to inquire about a user's friendliness.
         '''
+        if id in self._logins:
+            id = self._logins[id].id
+
         return id in self.friends
 
     def isFoe(self, id):
         '''
         Convenience function for other modules to inquire about a user's foeliness.
         '''
+        if id in self._logins:
+            id = self._logins[id].id
+
         return id in self.foes
 
-    def isPlayer(self, name):
+    def isPlayer(self, id):
         '''
         Convenience function for other modules to inquire about a user's civilian status.
         '''
-        return name in self or name == self.me.login
+        return id in self
 
     def getUserColor(self, id):
         '''
@@ -109,12 +115,12 @@ class Players:
     def __getitem__(self, item):
         if isinstance(item, Player):
             return item
-        try:
-            # Lets hope that nobody has an integer valued name
-            return self._players[int(item)]
-        except (ValueError, KeyError):
-            if item in self._logins:
-                return self._logins[item]
+        elif item in self._players:
+            # item is a player id
+            return self._players[item]
+        elif item in self._logins:
+            # item is a player name
+            return self._logins[item]
 
     def __setitem__(self, key, value):
         assert isinstance(key, int)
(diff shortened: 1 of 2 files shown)


src/client/players.py
```diff
@@ -45,19 +45,25 @@ def isFriend(self, id):
         '''
         Convenience function for other modules to inquire about a user's friendliness.
         '''
+        if id in self._logins:
+            id = self._logins[id].id
```
ChrisKitching, line 49: Wait, what? If logins is indexed by user id, what's the point of using it to look up user id?
Are there two different coexisting ID schemes or something (if so, why?)

HardlySoftly, line 49: The coexisting lookup scheme is the player name, so id might be a user name string.

The code is still transitioning from using names everywhere into using IDs instead, but until that's done I feel like the isFriend and isFoe methods should support the old way of doing things.

Sheeo, line 49: It's almost done, and I do believe friends are queried entirely by id.

The current `players['Name']` implementation does do fallback lookup by strings, albeit not in the most efficient manner. It should be fine for now though.

ChrisKitching, line 49: Why not just make this code require IDs and just fix up the slew of ensuing crashes? seems like the quickest way to complete the migration.

Sheeo:
See @ChrisKitching's comment

HardlySoftly:
Dropping support for look-ups by name is a) not always possible and b) one of the easiest ways I can think of to introduce subtle bugs that I miss in my limited testing time.

The last thing I think we want are bugs introduced accidentally, then taking 6 months for someone to come around and fix them in 5 mins.  This has happened a lot more than is ideal, and I endeavour to not cause these kinds of problems.

This commit isn't supposed to legitimise the use of look-ups by name, it simply recognises that it already happens (the last time I measured it, there were nearly 8000 look-ups by name when I started my client).

Sheeo:
Why isn't it always possible? In cases where it isn't possible we should make it so. For instance when loading replays; we should be storing player id's in addition to player names.

I'm aware there are a few server messages that still carry only a name; we need those sorted as well.

The subtle bugs will always be accompanied by the FIXME message that was previously emitted during lookup by name -- so we should at least ensure this stays around.

HardlySoftly:
I recently got the client to do a bunch of lookups by name in the case that IRC chat list arrives before player list.

Also https://github.com/FAForever/client/issues/352 is another example

ChrisKitching:
Use of python 3.5 type annotations will make the introduction of subtle bugs like this basically impossible in this case.

> I recently got the client to do a bunch of lookups by name in the case that IRC chat list arrives before player list.

It seems to make more sense to, when this happens, just wait for the player list to arrive. With model/view stuff this should be very pleasant to do.

More generally, we just can't use names. Because of the way we allow name changes, they're not just slow, they're _wrong. It's something we should work towards eliminating, although I get that we probably need to continue support for the moment.

HardlySoftly:
I'm certainly not endorsing this type stupidity, just want to explicitly point out it's existence.

ChrisKitching:
Excellent. Let's keep making progress, then.

PR #362, opened 2016-06-08
Fixes for bad usage of 'players' methods

HardlySoftly:


src/chat/channel.py
```diff
@@ -308,10 +308,8 @@ def printRaw(self, name, text, scroll_forced=False):
         '''
         Print an raw message in the chatArea of the channel
         '''
-        if self.lobby.client.players.isPlayer(name):
-            id = self.lobby.client.players[name].id
-        else:
-            id = -1
+        id = self.lobby.client.players.getID(name)
```
ChrisKitching, line 311: I'm thinking we may want to make the players thing global at some point: these parent pointers are gnarly as hell.

Perhaps we'll even want to go all model/view on this and make the players system a Qt Model at some point.

src/client/players.py
```diff
@@ -102,6 +102,11 @@ def get(self, item, default):
         val = self.__getitem__(item)
         return val if val else default
 
+    def getID(self, name):
+        if name in self._logins:
+            return self._logins[name].id
+        return -1
```
ChrisKitching, line 108: This should definitely log something if it can't find an ID. The current behaviour will cause an obscure crash in whatever called getID when the -1 is eventually used, without leaving a nice easy to follow trail.

I suggest logging something at the WARN level. Or maybe even throw an exception?

src/chat/channel.py
```diff
@@ -305,31 +308,32 @@ def printRaw(self, name, text, scroll_forced=False):
...
-            formatter = Formatters.FORMATTER_RAW
-            line = formatter.format(time=self.timestamp(), name=name, color=color, width=self.maxChatterWidth, text=text)
-            self.chatArea.insertHtml(line)
-            
-            if scroll_needed:
-                self.chatArea.verticalScrollBar().setValue(self.chatArea.verticalScrollBar().maximum())
-            else:
-                self.chatArea.verticalScrollBar().setValue(scroll_current)
-        except:
-            pass
```
ChrisKitching, line 332: `except: pass`

Whaaaaaaat.

ChrisKitching:
Great work. I'll add the log message I mentioned and merge this.

HardlySoftly:
I agree in principle, I would just check it doesn't cause log spam from IRC players being checked all the time.

HardlySoftly:
So I checked what happens with the logging of -1 returns, and the answer was 443 warnings on client startup

HardlySoftly:
Without this logging, we already have a significant improvement on the previous code behaviour.

What really needs to be done next, is to have an extra method for checking if some name is a player (or otherwise assume irc), before we fetch the player id.

The current usage of getId assumes it may return -1, but I don't think this is behaviour we want to encourage.

ChrisKitching:
This managed to fall out of my to-do list. I will merge it now without the log statement. Thank you.

ChrisKitching:
Merged to develop.

PR #364, opened 2016-06-10
Use the development toggle in URLs

aeoncleanse:
Not really sure what's going on in downloadmanager and tutorialitem with those huge diffs. Something to do with line formatting?


heelix:
For those of us trying to set up a local dev environment, anyone got a copy of the /etc/hosts file?

aeoncleanse:
There we go, dealt with the line formatting, and I've split it into two sensible commits.

src/chat/chatter.py
```diff
@@ -295,11 +296,13 @@ def pressed(self, item):
                 action_inspect_in_mordor = QtGui.QAction("Send the Orcs", menu)
                 menu.addAction(action_inspect_in_mordor)
 
+                ROOT = Settings.get('mordor/host')
+
                 def send_the_orcs():
                     if self.id != -1:
-                        QtGui.QDesktopServices.openUrl(QUrl("http://mordor.faforever.com/users/{}".format(self.id)))
+                        QtGui.QDesktopServices.openUrl(QUrl("http://" + ROOT + "/users/{}".format(self.id)))
```
Sheeo: Why not

``` python
"http://{}/users/{}".format(Settings.get('mordor/host'), self.id)
```

?

aeoncleanse: Oh cool, that works? Probably better do that then! This will work for all the instances I've changed in this PR?

ChrisKitching: I guess this means you don't know how the format function works. It basically does substitutions: each argument to the format function replaces the corresponding "{}" in the string. You can have as many as you like. It converts each argument to a string using that object's `repr` function, I believe.

You can do an even neater thing in python:

`"Here are some numbers: %s, %s, %s" % (1, 2, 3)`

Is equivalent to `"Here are some numbers: 1, 2, 3"`.

You might want to do it that way here.

That said, an even better way is to hide this `ROOT` rubbish. Why does every single usage of any of the FAF web routes have to do this string building itself? That's annoying. Wouldn't it be nice to be able to do something like:

`QtGui.QDesktopServices.openUrl(getRoute("/users/%s", self.id))`

That should be rather obviously doable, I hope?

aeoncleanse: All the format stuff makes sense. Your final comment, though... No, not obvious. I think I get what you mean, and how it works in this case though. Is using %s better than using {} and format? Then write a function which grabs the target from config and puts it at the start?

yorick-ne: https://pyformat.info/ for some documentation 
Also this style `"%s% %("test")` might not work in python 3 ( not 100% sure on this )

Sheeo: As a style note, `"{}".format(foo)` should be preferred over the classic `"%s" % foo`, since it's less error-prone and the format syntax is more powerful. You can later rearrange presentation without rearranging order of arguments, it supports aligning when padding, and more easily readable keyword arguments:

``` python
data = {'first': 'Hodor', 'last': 'Hodor!'}
# Old
'%(first)s %(last)s' % data
# New
'{first} {last}'.format(**data)
```

Sheeo:
Actually, I suggest we change the defaults and do something like this post-config default loading ([Here](https://github.com/aeoncleanse/client/blob/76ea257b8a59e4dcdf0a0c13cd81169929b33739/src/config/__init__.py#L122)):

``` python
for k,v in defaults.items():
    if isinstance(v, str):
        defaults[k] = v.format({'host', Settings.get('host')})
```

Configuration values can then be changed to contain `{host}` and it will be automatically substituted with, at preference, what's in your local .ini host key, and fallback to `default['host']` which would be set to `test.faforever.com` for development and just `faforever.com` for production.

While we're at that, it should output the main host used.

aeoncleanse:
Sorry, that just went way, way over my head. I don't remotely understand what you're getting at. Each of these files still has to import from settings at some point, no?

ChrisKitching:
Merged d8ff894

PR #365, opened 2016-06-12
Do not create extra/sounds/voice

micheljung:
Voices are now downloaded and mounted as .nxt files, so there is no need for these directories

### DIFF
--- a/src/fa/maps.py
+++ b/src/fa/maps.py
@@ -17,7 +17,7 @@ import zipfile
 import tempfile
 import re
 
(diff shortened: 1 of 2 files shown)


ChrisKitching:
Why was this ever necessary? Can't maps just provide sounds in their zipfiles anyway that get picked up when the map itself is mounted in the VFS?

micheljung:
It was never necessary, that's why we removed it.

But we don't packe the sounds which are pretty big) into the maps since then they have to be packed, uploaded, stored and downloaded again and again if the map changes.
As long as we don't have efficient patching, this is undesired.

Issue #351, opened 2016-05-30
Horrible performance in replayswidget


HardlySoftly:
Specifically this appears to be a really expensive operation: https://github.com/FAForever/client/blob/develop/src/replays/_replayswidget.py#L441

It gets called for every player in each gameInfo item that the client receives, which on startup incurs can be a lot.

For reference when I tested this line was being called ~1300 times and was adding a startup delay of 11 seconds to the client (more time than everything else put together)

Crotalus:
Should probably use `setUpdatesEnabled(false)`  when batch updating stuff like that, probably same thing when populating nicknames in #aeolus

Wesmania:
Is this still an issue?

PR #375, opened 2016-06-19
Remove slow code

HardlySoftly:
Solves https://github.com/FAForever/client/issues/351

### DIFF
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -438,7 +438,6 @@ class ReplaysWidget(BaseClass, FormClass):
                         playeritem.setDisabled(True)
 
                     item.addChild(playeritem)
-                    self.liveTree.setFirstItemColumnSpanned(playeritem, True)
         elif info['state'] == "closed":
             if info['uid'] in self.games:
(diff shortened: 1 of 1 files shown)


ChrisKitching:
Don't we need to call `setFirstItemColumnSpanned` once somewhere? I'm not 100% sure what it does, even after glancing at the manual. This was presumably added for a reason. Hopefully.

Would be nice if this really was all we need to do...

HardlySoftly:
Yeah, if you look in the live replays tab after this change, you'll see some bevels in the middle of the live replay things.  This call merges all the underlying columns together to remove this, at the cost of being easily slower than the rest of the client startup combined.

Kinda minor UI thing, which we can change later (it is already like this in local replays for example)

ChrisKitching:
> This call merges all the underlying columns together to remove this, at the cost of being easily slower than the rest of the client startup combined.

I find it difficult to believe that there isn't a faster way of doing this (bonus points for working out what it is ;) ), but for now am happy to have this slight UI derp to make the client usable again. Good job.

Especially since it's in a rarely-used tab. I'm working on some stuff that should get us away from this "initialise the universe and then never touch any of it" crap we currently do.

Issue #376, opened 2016-06-21
New avatars don't work


speed2CZ:
New avatars that got uploaded are not properly shown as you can see in the picture for Ladder Warrior avatar.
![utpdvyx](https://cloud.githubusercontent.com/assets/10947024/16225133/7b00c89c-37a5-11e6-83ce-1e9962281b73.png)

If given to player, pretty much same result in avatar select window
![ss 2016-06-21 at 11 45 53](https://cloud.githubusercontent.com/assets/10947024/16225207/cbfe8306-37a5-11e6-8297-17ee3031c47d.png)

When I select that avatar and restart client, I get an error. Client says that the bug was automatically reported.
Log here: https://gist.github.com/speed2CZ/42b605b18590bc75d64e8a2c22e96ca5

P.S. These avatars work in Downlord's client

HardlySoftly:
The bug is here:
2016-06-21 11:46:53,522 ERROR    **main**                       Uncaught exception
Traceback (most recent call last):
  File "src\chat_chatwidget.py", line 127, in finishDownloadAvatar
TypeError: 'NoneType' object is not iterable

I'll take a look

HardlySoftly:
So the server was sending encoded URLs to the client in certain cases (I dunno why) in the player info messages.

Meanwhile when the avatars were stored in src/util, they were being indexed by the decoded URL (this was automatically being handled by QUrl)

That meant there was a mismatch between the key for the avatar file, and the key that was being used to try and find it.

Sheeo:
Server was sending encoded URL's, which has been changed back now for the ones that were affected.

Client still needs to handle this properly, so thanks for the pr :)

HardlySoftly:
I think I will look at refactoring the avatar code since the PR is a bit of an ugly hack.

PR #378, opened 2016-06-21
Decode avatar URLs that the server sends

HardlySoftly:
Fixes https://github.com/FAForever/client/issues/376


ChrisKitching:
Excellent work.

PR #382, opened 2016-06-26
Queue up bind requests if game conn isn't ready

Crotalus:
Seems like peer_bound requests sometime gets triggered before the FA process has connected to the client (thus no game_connection active).

Example of user experiencing this error:
http://forums.faforever.com/viewtopic.php?f=3&t=12577

```
2016-06-11 22:45:21,256 INFO     connectivity.relay.Relay       Allocating local relay for turinturambar, 136480
2016-06-11 22:45:21,256 INFO     fa.game_session.GameSession    Bound peer turinturambar/136480 to 65442
2016-06-11 22:45:21,256 ERROR    __main__                       Uncaught exception
  File "src\fa\game_session.py", line 123, in _peer_bound
AttributeError: 'NoneType' object has no attribute 'send'
2016-06-11 22:45:21,365 INFO     requests.packages.urllib3.connectionpool Starting new HTTP connection (1): api.dev.faforever.com
2016-06-11 22:45:25,678 INFO     fa.game_session.GameSession    Game connected through GPGNet
```

@sheeo, is it theoretically possible that a slow FA startup can cause this?

If this is the case then this PR will fix that by queue up peer_bound requests which then is executed when the actual game connection is established (maybe there is a better way to solve this?)

### DIFF
--- a/src/fa/game_session.py
+++ b/src/fa/game_session.py
@@ -1,7 +1,7 @@
 from PyQt4.QtCore import QObject, pyqtSignal
 from PyQt4.QtNetwork import QTcpServer, QHostAddress
 from enum import IntEnum
-
+from functools import partial
 from base import Client
 from connectivity.turn import TURNState
 from decorators import with_logger
@@ -56,6 +56,8 @@ class GameSession(QObject):
         # Use the normal lobby by default
         self.init_mode = 0
         self._joins, self._connects = [], []
+        # queue up bind requests if game connection isn't initialized
+        self._bind_queue = []
 
         # 'GPGNet' TCP listener
         self._game_listener = QTcpServer(self)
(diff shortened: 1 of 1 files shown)


Crotalus:
Happens in 1v1 ladder when the ForgedAlliance.exe of player flagged HOST is slow to startup, in the above example the GpgNet connection connects to client 4 seconds after it receives the `JoinGame` command from server.

The problem is that when the joining client sends 'Lobby', server will send out command to both the host / joining player regardless if the host is ready (happens in `ConnectToHost`):
https://github.com/FAForever/server/blob/develop/server/gameconnection.py#L130

I think this needs to be handled server side instead, open for ideas how to solve this the best way (or shoot down my observations).

Crotalus:
Resubmitting a server solution

PR #384, opened 2016-07-07
use the new uid executable

muellni:


muellni:
I tested the setup from https://ci.appveyor.com/project/Sheeo/client/build/589/artifacts and it works. The `uid.exe` creates a popup console window on Login, which I fixed in https://github.com/FAForever/uid/releases/tag/v2.1.0

ChrisKitching:
We probably don't want to check the binary into this repository, but instead to use some sort of build system and pull in the other repo as a submodule and trigger a build of it.

muellni:
The uid executable is downloaded [via travis](https://github.com/muellni/client/blob/d172fc661cd8cfdaccb91f04c2faffb2db968d80/.travis.yml#L13) and [via appyevor](https://github.com/muellni/client/blob/d172fc661cd8cfdaccb91f04c2faffb2db968d80/appveyor.yml#L22), similar to the python module before.

muellni:
replaced by #453

PR #385, opened 2016-07-10
Remove double dependency

aeoncleanse:


ChrisKitching:
#386 does this and also fixes appveyor

PR #395, opened 2016-08-08
Move ConnectivityHelper to separate thread

### LINKED ISSUE #348: Slow startup + traffic blocked
Last few lobby releases this happens: FAF opens then freezes for 10-40 seconds. Can't click anything. The longer it takes the higher chance I'll get traffic blocked.
https://gist.github.com/speed2CZ/edd0c39eb34428a68c2041a96f942824

Crotalus:
Seems to be a good idea to move ConnectivityHelper to a separate thread .This way we don't care if the GUI locks up and it will also reduce jitter / latency when relaying the in-game packets.

Hopefully fixes #348 (Outbound traffic error) et al.

(needs testing, I'll run a couple of games later this week with this patch)

### DIFF
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -1,6 +1,6 @@
 from functools import partial
 
-from PyQt4.QtCore import QUrl
+from PyQt4.QtCore import QUrl, pyqtSignal
 from PyQt4.QtGui import QLabel, QStyle
 from PyQt4.QtNetwork import QAbstractSocket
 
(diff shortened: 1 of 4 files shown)


duk3luk3:
Sounds like a very important patch we should try to get in. Can you get this ready @Crotalus ?

duk3luk3:
How does QThread interact with the python GIL?

Crotalus:
Hmm this branch should be seen as deprecated now that we soon have the standalone ice adapter.

As Sheeo told me earlier, spawning this off to a separate thread won't help that much due to GIL so I suggest we close this.

duk3luk3:
Well, you'd fork it off to a process... python has no problem doing that.

PR #402, opened 2016-08-26
Fix #397 - autodownload more then one mod when joining a game

### LINKED ISSUE #397: Auto mod download - DL only a single mod, then launches, causing prob with heavily modded games
Joining a lobby with multiple mods that you do not have, results in only 1 being downloaded, and then launching. This usually results in getting auto-kicked by the host, for not having all the mods. Forced to repeatedly join the lobby (mod vault is littered with outdated versions, so searching it is no viable alternative).

yorick-ne:


Crotalus:
:+1:

PR #406, opened 2016-08-31
Add a button to view ladder maps

yorick-ne:
Is still WIP, but this is what i got so far. The core functionality is there, but it probably needs polishing.

There are still a couple things to be done:
- [ ] FAForever/api#96 - the api returns the wrong maps
- [ ] use the faftools api client

There are currently 3 classes in the same file, dunno how good that is coding-style-wise.


duk3luk3:
I'm not entirely sure what the state is with ladder maps, but could you fix this up?

PR #413, opened 2016-09-23
Refined Play - Lobby info

GrotheFAF:
fixed oversized Map-Icons.
add mod info to Lobby-info.
fix html on games\formatter.
match look of replay-info and lobby-info.
rewrote replayitem-infoPlayers.
flltered out 'cop-maps'-error in fa\maps.
rewrote (without formatter) on stats\ def createResults.


Sheeo:
Please refine your commits into smaller, more descriptive ones, then I'll take a look.

Make sure you:
- Don't include whitespace/cleaning changes alongside semantic ones (This makes the diff unecessarily hard to read)
- Use [well formed commit messages](http://chris.beams.io/posts/git-commit/)
- Don't use the exact same commit message twice

requirements.txt
```diff
@@ -1,7 +1,6 @@
 bsdiff4==1.1.4
 cx_Freeze==4.3.4
 enum34
-lupa
```
Sheeo, line 4: Why have you removed this dependency? It's a runtime dependency of faftools and we're using it to parse lua code.

You left no comment about this in the PR, either.

src/games/gameitem.py
```diff
@@ -161,38 +157,34 @@ def announceHosting(self):
...
-        self.host = message['host']
+        # Updates this item from the message dictionary supplied #
 
-        if 'host_id' in message:
-            self.hostid = message['host_id']
-        else:
-            self.hostid = self.client.players.getID(self.host)
+        self.client = client
+
+        if self.title is None:  # you can't change this...
```
Sheeo, line 165: What does this comment mean, and why are we now only assigning these properties in update if `title is None`?

src/games/gameitem.py
```diff
@@ -204,6 +196,14 @@ def update(self, message, client):
             client.usersUpdated.emit(self.players)
             return
 
+        # Map preview code
```
Sheeo, line 199: Why was this moved up here?

src/games/gameitem.py
```diff
@@ -294,70 +292,66 @@ def update(self, message, client):
...
+        if len(observerlist) != 0:
             observers = "Observers : "
             observers += ",".join(observerlist)        
 
-        mods = ""
-
         if self.mods:
-            mods += "<br/>With " + "<br/>".join(self.mods.values())
+            mods = "<br />With: " + "<br />".join(self.mods.values())
+        else:
```
Sheeo, line 348: This branch is redundant since `mods` is initialized to `""`.

src/games/gameitem.py
```diff
@@ -294,70 +292,66 @@ def update(self, message, client):
...
+            mods = ""
 
-        self.setToolTip(self.FORMATTER_TOOL.format(teams = teams_string, observers=observers, mods = mods))
+        self.setToolTip(self.FORMATTER_TOOL.format(teams=teams_string, observers=observers, mods=mods))
 
     def permutations(self, items):
-        """Yields all permutations of the items."""
+        # Yields all permutations of the items. #
```
Sheeo, line 354: Please stick to standard python doc-comments. These are formatted using triple-quotes.

src/replays/replayitem.py
```diff
@@ -105,175 +90,194 @@ def __init__(self, uid, parent, *args, **kwargs):
...
 
     def infoPlayers(self, players):
-        '''
-        processes information from the server about a replay into readable extra information for the user, also calls method to show the information
-        '''
+        # processes information from the server about a replay into readable extra information for the user, #
+        # also calls method to show the information #
+
         self.moreInfo = True
+        self.numberplayers = len(players)
```
Sheeo, line 151: Certainly `len(players)` is more descriptive than an instancefield called `numberplayers`.

src/replays/replayitem.py
```diff
@@ -105,175 +90,194 @@ def __init__(self, uid, parent, *args, **kwargs):
...
-                if not team in self.teams :
+
+        for player in players:  # player highscore
+            if "score" in player:
+                if player["score"] > mvpscore:
+                    mvp = player
+                    mvpscore = player["score"]
+
+        for player in players:  # player -> teams & playerscore -> teamscore
+            if self.mod == "phantomx" or self.mod == "murderparty":
```
Sheeo, line 163: What is this?

Sheeo (changes requested):
Please submit updated commits without the extraneous formatting changes and focused on the UI changes you want made.

src/games/gameitem.py
GrotheFAF: Once you have a title, you are hosting and you can't change these properties from inside a lobby.

src/games/gameitem.py
GrotheFAF: long story .. I had a 'if not hidden' from L230-276 (but I took it out again, was not sure sure), and the mapname part was needed even when hidden, so I moved it outside the 'if not hidden' condition.

src/games/gameitem.py
GrotheFAF: I removed that (L352) and made the branch instead.

src/games/gameitem.py
GrotheFAF: ok. (pyCharm is complaining about the single triple-quotes)

requirements.txt
GrotheFAF: I was told during dev-school to remove it.

src/replays/replayitem.py
GrotheFAF: yes, sad story... I needed that number outside the players context. (in resize)

src/replays/replayitem.py
GrotheFAF: It's about putting ffa-like games in one team for the replay-info. 
(because in some replays of these mods the player are in one team and in others they are not, and I wanted a consistent display. And initially (this commit) to avoid the unpractical width of .. vs .. vs .. vs .. vs ..)

src/games/gameitem.py
Sheeo: Yet. There's nothing stopping us from adding that functionality. So why arbitrarily limit it here?

requirements.txt
Sheeo: Don't do that.

src/games/gameitem.py
Sheeo: Then move this back again where it's used, potentially inline it to make it less verbose?

src/games/gameitem.py
Sheeo: Woops. Indeed.

Would prefer something like this, but not important at all.

``` python
mods = "<br/>With " + "<br />".join(self.mods.values()) if self.mods else ""
```

src/games/gameitem.py
Sheeo: My PyCharm is not complaining about that. What error is it giving you?

src/replays/replayitem.py
Sheeo: Why not store `self.players` then?

src/replays/replayitem.py
Sheeo: That's quite a nasty hack. Let's not do that -- or at the very least, leave a comment about it?

requirements.txt
GrotheFAF: Was not my idea. I was just doing what I'm told.

src/replays/replayitem.py
GrotheFAF: tmi .. I don't need all that... just one number.

src/replays/replayitem.py
GrotheFAF: I know... but I didn't saw any alternativ to ffa these mods.
Even if I went the extra round to find out if len(teams) = len(players), it wouldn't cover all cases. (there are mixed cases with some players with team and others without.)

src/games/gameitem.py
GrotheFAF: I was trying to make this function more efficient, cause it is 'on' all the time. 
(and it doesn't limit.. it can be adapted if needed.. but I see the point if that outweight that some future functionality-adder might miss it and wonders why it's not working.)

src/games/gameitem.py
GrotheFAF: I'm not over the 'if not hidden' part, I plan to put it back in.
(and I don't understand the 'less verbose' part.)

src/games/gameitem.py
GrotheFAF: well, that looks foreign to me (and slower?) ... but then again ... I'm not from python land.

src/games/gameitem.py
GrotheFAF: not the -double- the triple-**single**-quotes ( ''' ).  (found for example below vv)

Sheeo: Ah yes, use `"""` for comments please, this is the most standard and it's used mostly throughout our python codebase.

src/games/gameitem.py
Sheeo: Slower in what sense?

It's python's version of the ternary conditional operator, in e.g. C-like languages:

``` javascript
var mods = mods ? mods.values() : "";
```

requirements.txt
Sheeo: Okay, please undo.

src/games/gameitem.py
Sheeo: I will bet you we aren't going to see any performance wins by not performing these assignments. There are way larger fish to fry and it'll just cause a headache later on when someone stumbles into this weird condition.

src/replays/replayitem.py
Sheeo: Okay, then at least call it `numberOfPlayers` (Though pascalCasing is inconsistently used throughout the codebase; please try to stick with underscores, e.g. `number_of_players`), or simply `playercount`.

PR #415, opened 2016-09-23
Replays tab UI change for performance gain

HardlySoftly:
This is what I got when profiling the code:
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    1794   20.571    0.011   20.571    0.011 {built-in method setFirstItemColumnSpanned}

This was when I started up the client, which took ~30 secs.

I already killed one of these calls (see https://github.com/FAForever/client/commit/3f5d946151ca634f8d8de383af91a0745d7cf0c8 ), but must have missed this one.

### DIFF
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -433,7 +433,6 @@ class ReplaysWidget(BaseClass, FormClass):
                         playeritem.setDisabled(True)
 
                     item.addChild(playeritem)
-                    self.liveTree.setFirstItemColumnSpanned(playeritem, True)
         elif info['state'] == "closed":
             if info['uid'] in self.games:
                 self.liveTree.takeTopLevelItem(self.liveTree.indexOfTopLevelItem(self.games[info['uid']]))


TAPAKAH4Life:
Thank god for people who do code improvements that matter

HardlySoftly:
So on reflection, it looks like I didn't miss this line of code last time, but it got re-added in another PR https://github.com/FAForever/client/pull/392

Sheeo:
@HardlySoftly that commit hasn't been released, so it can't be the cause of this.

GrotheFAF:
but PR #392 was the way it got back into develop, right? .. (that's how I 'got it')

Sheeo:
My point being, it cannot be the degradation that @TAPAKAH4Life is experiencing.

HardlySoftly:
Sheeo I only profiled the latest dev client, and I think its better if we don't make the client any slower if we can help it.

Is there any way to automatically profile the code when you make a PR (might be fun)?

PR #423, opened 2016-09-30
Refactor UPnP code and add support for NAT-PMP

Crotalus:
Add support for NAT-PMP and use that as primary way of requesting NAT before using w32-UPnP. Maybe also should look at have a second UPnP implementation for non-w32 + cases where it is incompatible with crappy routers.

miniupnpc seems like a good option in that case.


duk3luk3:
Do we still want this?

Crotalus:
AFAIK ice-adapter will handle upnp so lets close this for now

PR #427, opened 2016-10-10
Not load Leagues and focus on Ladder Ratings (+ beautify and cleanup)

GrotheFAF:
I hope the XP-exception doesn't offend anybody ... as it might be just for me.


Brutus5000:
This is a subset of #431. #431 got merged with #433, therefore I close this one.

PR #431, opened 2016-10-12
highlight the selected replay in list

GrotheFAF:


Brutus5000 (dismissed):
Commit 363982b has 2 possible breaks:
- [ ] check that the import-changes (remove of src) doesn't break anything
- [ ] check the changes regarding type-checks

Commit be5f447 changes png-metadata and remove old mod-info-html files. These should come via API now, so should do no harmful change.

Commit 8f3cd4d contains no harmful changes.
Mostly changes in code style. A few html-fixes (closing tags).

Commit 3500475 should work (I have no testing environment).

Commit 0b66c7c should be partially reverted.

Commit 24e2f37 works without drawbacks.

Obviously I can't review just **one single** commit... well I will keep this comment up to date.

src/secondaryServer/secondaryserver.py
```diff
@@ -111,14 +115,14 @@ def writeToServer(self, action, *args, **kw):
         out.writeUInt32(0)
         out.writeQString(action)
         
-        for arg in args :            
-            if type(arg) is IntType:
+        for arg in args:
+            if type(arg) is type(int):
```
Brutus5000: This is wrong in Python 3.x and 2.7. Example:

```
>>> type(5) is int
True
>>> type(5) is type(int)
False
```

Has this been tested?

src/secondaryServer/secondaryserver.py
```diff
@@ -111,14 +115,14 @@ def writeToServer(self, action, *args, **kw):
...
                 out.writeInt(arg)
             elif isinstance(arg, basestring):
                 out.writeQString(arg)
-            elif type(arg) is FloatType:
+            elif type(arg) is type(float):
```
Brutus5000: same as above

src/secondaryServer/secondaryserver.py
```diff
@@ -111,14 +115,14 @@ def writeToServer(self, action, *args, **kw):
...
                 out.writeFloat(arg)
-            elif type(arg) is ListType:
+            elif type(arg) is type(list):
```
Brutus5000: same as above

Brutus5000 (commented):
Regarding commit 3500475: Why is this not supposed to popup on XP? An explanation would be fine.

src/secondaryServer/secondaryserver.py
GrotheFAF: found no case with args during debug .. so not really tested.
but your right, it's wrong .. seems right at the time .. hmm.
how can it be fixed.. another fix-commit?

src/client/_clientwindow.py
```diff
@@ -1152,7 +1152,8 @@ def handle_invalid(self, message):
         raise Exception(message)
 
     def handle_stats(self, message):
-        self.statsInfo.emit(message)
+        # self.statsInfo.emit(message)  # currently useless -> safe some time
```
Brutus5000, line 1155: The time safed does not compete with the time lost, when a developer relies on the qt signal and tries to find the error...
If there is no significant slowdown, we should keep it.

GrotheFAF: it is significant .. loading the many thousand users for no purpose (the league is 'out of order'), when you want to have a look at ladder ratings.

GrotheFAF:
in XP it runs with admin rights, so without the exception faf stops with that error.

Brutus5000:
Regarding commit 804176f: Inlining html back to python code is no enhancement to me. I would skip this commit.

src/client/_clientwindow.py
Brutus5000: Your comment explains the change in https://github.com/FAForever/client/pull/431/commits/0b66c7c15f96b76bf0992a011e251a389bfac3ff#diff-18b50d81a733c795841561309848bd40R202.

But handle_stats is called only, when the server sends stats.

src/secondaryServer/secondaryserver.py
Brutus5000: Yes.

src/client/_clientwindow.py
GrotheFAF: ok, it sends stats ..my change only prevents the displaying process.
so I might comment too: https://github.com/FAForever/client/pull/427/commits/0b66c7c15f96b76bf0992a011e251a389bfac3ff#diff-18b50d81a733c795841561309848bd40R203
no more sending... but that needs more study...

src/secondaryServer/secondaryserver.py
GrotheFAF: should I branch this fix/commit on develop?

GrotheFAF:
well, ok ... commit 804176f  was more of a rehersal of append/join. (which I read is much much faster than str+= ....)

src/secondaryServer/secondaryserver.py
Brutus5000: No, put it into GrotheFAF:feature/replays-highlight . This way it get's added to this PR automatically.

GrotheFAF: ok done.

Brutus5000:
Merged with #433, therefore I close this one.

PR #435, opened 2016-10-19
Feature/replays automatic joinlive

### LINKED ISSUE #414: search replays by player
when u search replays by player and switch back and forth to the replay tab, the name of player remains in the search field, but the replays are reloaded and not specified anymore

GrotheFAF:
add checkbox to disable automatic-refresh Replays (fixes https://github.com/FAForever/client/issues/414)
show live games as 'Playing' in Replays
fix duration for live games in Replays
adds join live games in Replays


Brutus5000:
I am unhappy with the wording for the new checkbox. The user does not know what "automatic" means or in which cases the replay vault get's refreshed at all.

duk3luk3:
> I am unhappy with the wording for the new checkbox. The user does not know what "automatic" means or in which cases the replay vault get's refreshed at all.

I'd say there shouldn't be a checkbox, just disable the auto-refresh when the user does a search, and add a button to "clear" the search.

GrotheFAF:
that is not exactly what this is about .. it's about the automatic refresh of recent games when you change tabs (out and back into Replays) .. if you had a search you like to keep ... you mark 'keep search' and you can change tabs without the search disapearing with a refresh of recent games if you reenter Replays.

duk3luk3:
No, that's exactly what I mean. When the user makes a search (clicks the search button) a "Clear search" button should appear and the refresh when switching tabs should be disabled. When the user clicks the "Clear search" button the search is refreshed to the standard view and the refresh when switching tabs is enabled again.

duk3luk3:
I like that better, but "Automatic" doesn't intuitively convey what the behaviour of the tab is when the checkbox is selected or not.

I don't think anybody wants the "Refresh when changing tab" behaviour at all; if there are users that want it you could convince me there's a point to having the checkbox but otherwise, I'd like to see it gone and have the behaviour that the only time the search is executed automatically is when the tab is selected for the first time (in the client session).

Of course anybody is free to disagree with me and make arguments... @Brutus5000 your opinion?

Brutus5000:
Without knowing how it worked I would have expected that the system keeps my search until I press clear.
My sugesstion would be to add a "keep search" checkbox which is enabled on default.

If anyone wants to get the old behavior back, he needs to deselect that checkbox,

duk3luk3:
OK, if both of you prefer the checkbox to the button, we'll do the checkbox.

GrotheFAF:
My last revision is back to a 'reserved' version of 'automatic', with the default to off, which means that user who want it the old way have to check it. (setting is preserved)

duk3luk3:
Sounds good, please rebase.

GrotheFAF:
rebased

duk3luk3:
The "automatic" checkbox seems to always be disabled now... what's up with that?

GrotheFAF:
In my part of the world it is working as expected. Please check again..  
(there was one state in commit history, where the checkbox was disabled during init.)

duk3luk3:
ok, the checkbox works, but it's always grayed out for me. Things should only be grayed out if they're disabled. I think it would also be better if it was labelled "Automatic Refresh" instead.

GrotheFAF:
you wanted it gone .. now it's nearly gone (=nearly visible) and now that is not ok either .. come on, man, make up your mind.

PR #436, opened 2016-10-20
speedup Leaderboards (and client start)

GrotheFAF:
rewrite createResults with append/join and use QTextBrowser insteat of QTextDocument
not load league on startup
fix an error with me.division


Brutus5000:
As stated in one of your recent PRs before, I will not merge code, where HTML code is inlined back into Python. I don't care whether this is faster, it is just bad style.
You may try to convince Sheeo or Crotalus though.

GrotheFAF:
I made a new version which kept the 2 formatter... just for you :-)

PR #437, opened 2016-10-22
make showprivateGames check visible

GrotheFAF:


Brutus5000:
Please describe your intended changes. As far as I can see, only the 1px border around the checkbox itself has changed it's color slightly.

GrotheFAF:
the background of the checkbox is now dark .. so you can actually see, when it is checked (text color = check color is white). (not an issue in W10)

Brutus5000:
I can't confirm your problems on a Windows XP-Installation either.
The problem must be on your local pc.

GrotheFAF:
my w7 shows the same problem. maybe it is me.... but I remember someone else complaining about that too..

Sheeo:
I don't understand the purpose of this either. Please clarify.

GrotheFAF:
I can't see the 'checked' in the checkbox (the background is white and the 'checked' is white too).
This will change the background to black[ish], so that I can see the white 'checked'.
(this might be different on other PCs, because I don't use any windows design (XP & W7)..simple windows classic)

duk3luk3:
I can confirm the issue in linux.

duk3luk3:
And I can confirm that this PR fixes it. Someone who has themes in their faf client needs to test that it plays nice with themes; or maybe the fix should be placed in the default theme.

GrotheFAF:
Yes, there might a beauty-problem with themes, cause the background of the textbox will be no longer TransParent.

duk3luk3:
I'm going to merge this so it doesn't keep languishing, but someone who has (client) themes installed really needs to check the impact on themes.

PR #438, opened 2016-10-22
speedup Leaderboards (and client start) + keep formatter

GrotheFAF:
rewrite createResults with append/join and use QTextBrowser insteat of QTextDocument
not load league on startup
fix an error with me.division

### DIFF
--- a/res/stats/formatters/ladder.qthtml
+++ b/res/stats/formatters/ladder.qthtml
@@ -1,5 +1,5 @@
-<tr class="{type}">
-    <td class="rank">{rank}</td>
-    <td class="login">{name}</td>
-    <td class="rating">{score}</td>
+<tr bgcolor="{color}">
+    <td align="center">{rank}</td>
+    <td>{name}</td>
+    <td align="center">{score}</td>
 </tr>
--- a/res/stats/formatters/ladder_header.qthtml
+++ b/res/stats/formatters/ladder_header.qthtml
@@ -1,7 +1,6 @@
-<tbody>
-<tr>
-    <th class="rank" width="50">{rank}</th>
-    <th class="login" width="100%">{name}</th>
-    <th class="rating" width="50">{score}</th>
+<tr bgcolor="{color}">
+    <th align="center" width="50">{rank}</th>
+    <th width="100%">{name}</th>
+    <th align="center" width="50">{score}</th>
 </tr>
-</tbody>
+
--- a/src/stats/_statswidget.py
+++ b/src/stats/_statswidget.py
@@ -1,17 +1,14 @@
 
 
-
-
 from PyQt4 import QtCore, QtGui, QtWebKit
 import util
 from stats import mapstat
+from config import Settings
 import client
 import time
 
 import logging
 logger = logging.getLogger(__name__)
-
-from config import Settings
 
 ANTIFLOOD = 0.1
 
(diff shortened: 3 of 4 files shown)


res/stats/formatters/ladder.qthtml
```diff
@@ -1,5 +1,5 @@
-<tr class="{type}">
-    <td class="rank">{rank}</td>
-    <td class="login">{name}</td>
-    <td class="rating">{score}</td>
+<tr bgcolor="{color}">
+    <td align="center">{rank}</td>
+    <td>{name}</td>
+    <td align="center">{score}</td>
```
Brutus5000, line 4: Why do you remove the classes and make changes hard-coded instead of putting it into the CSS?

res/stats/formatters/ladder_header.qthtml
```diff
@@ -1,7 +1,6 @@
-<tbody>
-<tr>
-    <th class="rank" width="50">{rank}</th>
-    <th class="login" width="100%">{name}</th>
-    <th class="rating" width="50">{score}</th>
+<tr bgcolor="{color}">
+    <th align="center" width="50">{rank}</th>
+    <th width="100%">{name}</th>
+    <th align="center" width="50">{score}</th>
```
Brutus5000, line 4: Same question as for ladder.qthtml

src/stats/_statswidget.py
```diff
@@ -93,8 +90,8 @@ def divisionUpdate(self, index):
     def createDivisionsTabs(self, divisions):
         userDivision = ""
         me = self.client.me
-        if me.division is not None:
-            userDivision = me.division
+        if me.league is not None:  # was me.division, but no there there
```
Brutus5000, line 93: The second part of the comment makes no sense.

src/stats/_statswidget.py
```diff
@@ -126,35 +123,28 @@ def createDivisionsTabs(self, divisions):
...
-        cursor = table.textCursor()
-        cursor.movePosition(QtGui.QTextCursor.End)
-        table.setTextCursor(cursor) 
-        color = "lime"
-        line = formatter_header.format(rank="rank", name="name", score="score", color=color)
-        html += line
+        glist = []
+        append = glist.append
+        append("<table style='color:#3D3D3D' cellspacing='0' cellpadding='4' width='100%' height='100%'><tbody>")
+        append(formatter_header.format(rank="rank", name="name", score="score", color="#92C1E4"))
```
Brutus5000, line 131: Again a class definiton that vanishes and hard-coded styling.

res/stats/formatters/ladder.qthtml
GrotheFAF: because QTextBrowser doesn't support stylesheets

duk3luk3:
@GrotheFAF look at this change I made on top of this PR: https://github.com/duk3luk3/client/commit/03c57eda85328e0a3ba308c65546ac29d9de8871

It shows that `<style>` works.

src/stats/_statswidget.py
duk3luk3: @Brutus5000  "There's no there there", makes perfect sense to me - but is a superfluous comment and should be deleted

duk3luk3:
OK, I've given it a more in-depth review and I am very happy with the code. Please just fix the following to get the PR into an acceptable state for merging:
- Ideally, cosmetic changes (like removing useless comments or fixing code formatting) should be in a separate commit, but if it's too much work to separate that now, you can leave it in this one (e.g. https://github.com/FAForever/client/pull/438/files#diff-18b50d81a733c795841561309848bd40R68)
- Do not comment out code you don't want to use anymore, just remove it (e.g. https://github.com/FAForever/client/pull/438/files#diff-18b50d81a733c795841561309848bd40R193)
- Rebase against develop
- Write a better commit message that lists all the changes you made (Replacing QTextEdit, Late-loading division tab, cleaning up code style) and include "Closes #460" in it

I am personally not hugely attached to having `<style>`, but if the stylesheet could be kept, it would be easier to change the style of the widgets later without going through the code and finding all the places you've put HTML tags with styling. That's the main reason stylesheets are used, not dogma ;-)

PR #440, opened 2016-10-27
fix selectReplay bugs

GrotheFAF:
'Spoiler Free' click produced an error when there was a selectedReplay before update onlineTree
'Spoiler Free' click produced an error when a date (instead of  Replay) was select


Brutus5000:
Please rebase onto develop or merge the latest develop into your branch. Currently it can't be merged with develop automatically.

PR #444, opened 2016-11-01
fix download Map Msg Title

GrotheFAF:


duk3luk3:
Straightforward fix :+1:

PR #453, opened 2016-11-05
use the new uid executable

muellni:
Replace `uid.dll` by `faf-uid.exe` provided by https://github.com/FAForever/uid


duk3luk3:
Merged in ee687a9

PR #467, opened 2016-11-09
Fix #466

muellni:
this commit basically cherry picks 7ff9ddd1072924c5877657a1486a06bd35e6d4a1 to fix #466

PR #468, opened 2016-11-09
ice adapter integration

muellni:
This is WIP for now and only connects to `test.faforever.com` which is currently "ICE-enabled".


muellni:
Is anyone able to make appveyor work? I'd like to test the MSI package.

duk3luk3:
appveyor fails sometimes, I think you need to restart the builds by amending your commit and force-pushing so github thinks it's a new commit

muellni:
So appveyor and travis now fail at the same spot: `tests/unit_tests/client/test_player.py::test_player_equality` is not terminating. 
Maybe related to this line: https://github.com/muellni/client/blob/61e8d4ceb09c61049fd8f1c1f1e0679710c18a0d/src/client/__init__.py#L19
I can't reproduce the tests on my machine, because something else fails: https://gist.github.com/ae7fdfe26df742fea8c3f6cb4c24c0eb

duk3luk3:
I think you're running the wrong environment or need to change the environment config, it should be skipping the `test_updater_is_a_dialog` test.

duk3luk3:
Also `WC1 /home/sws/projPriv/2016/faf-client-ice-adapter/tests/fa/test_featured.py cannot collect test class 'TestObjectWithoutIsFinished' because it has a __init__ constructor` shouldn't happen, I renamed those classes. Did you revert that with your fixup commits?

EDIT: git blame suggests my commits to fix that never made it into develop, so I'll have to check that myself.

muellni:
It must be some problem is this commit, since the parent passed all CI checks: https://github.com/FAForever/client/pull/467

duk3luk3:
Please rebase so we can test this this week

muellni:
K, here you go. It should work, but `src/connectivity/JsonRpcTcpClient.py` is just a stub and needs a proper implementation.

PR #479, opened 2016-11-12
3.5 refactoring

Crotalus:
Rebased my earlier 3.5 work to the latest develop, haven't tested it yet though due to lack of time and not having uid.exe.


duk3luk3:
will this still be python2-compatible?

Sheeo:
@duk3luk3 python2 and python3 are mutually incompatible.

While you can write code that runs on both versions, that's a huge maintenance burden and not really worth it for an application such as this.

duk3luk3:
Yeah, I'm aware of that.

duk3luk3:
Tried to pull this, after fixing some small things it starts up, but doesn't show chat or news, but if I click on games it shows 

`QAbstractSocket::connectToHost() called when already looking up or connecting/connected to "lobby.faforever.com"`

duk3luk3:
It's a string encoding error... and the general practice of the client to eat exceptions and just spit them into the log is terrible.

duk3luk3:
My branch now here (WIP): https://github.com/FAForever/client/tree/474-feature-py3 (pushed to this repo to trigger tests)

duk3luk3:
We need a new lupa build for appveyor I guess. @Sheeo how is that done?

Wesmania:
#596 seems to have done something with python wheels, and we'll probably need python3 from bireus - maybe we can rebase this and see if lupa isn't broken anymore?

Wesmania:
I did a little rebase - it seems unfortunately that trueskill python module is only compatible up to python 3.3 (at least that's what available at pypi).

Issue #470, opened 2016-11-09
firewall rules for faf-ice-adapter.exe


muellni:
Create Windows firewall rules to allow all incoming connections to faf-ice-adapter.exe during installation

muellni:
I'll resolve this by triggering the firewall dialog on client startup, so it doesn't become hidden behind the game

sikoragmbh:
The client now triggers the `reserveRelays` method, which triggers the firewall dialog after logging in with the client. This should be sufficient.

muellni:
The client now triggers the `reserveRelays` method, which triggers the firewall dialog after logging in with the client. This should be sufficient.

PR #484, opened 2016-11-14
470 fix installer firewall

muellni:
Fix for #470


muellni:
Not working yet...

muellni:
Still can't get it working

muellni:
https://sourceforge.net/p/cx-freeze/mailman/message/24474819/ says I'm doing it wrong: bdist_msi command in distutils is unable to run `install-script`.

muellni:
I'm a little bit afraid, that the bdist_msi installer option `--install-script` is not working at all or not with the version we use. 
You can create a `fafinst.log` log file using `msiexec /i "Forged Alliance Forever-0.11.60-win32.msi" /L*v fafinst.log`

PR #495, opened 2016-11-17
fix 'View Replays in Vault' change to 'Online Vault' Tab in Replays

GrotheFAF:
fixes (2nd point in) #414

- [x] Code has tests

- [x] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -384,6 +384,7 @@ class Chatter(QtGui.QTableWidgetItem):
     @QtCore.pyqtSlot()
     def viewVaultReplay(self):
         """ see the player replays in the vault """
+        self.lobby.client.replays.setCurrentIndex(2)  # focus on Online Fault
         self.lobby.client.replays.mapName.setText("")
         self.lobby.client.replays.playerName.setText(self.name)
         self.lobby.client.replays.minRating.setValue(0)
--- a/changelog.md
+++ b/changelog.md
@@ -37,7 +37,8 @@
     - Add Button 'Refresh Recent List'/'Reset Search to Recent'
     - Add Checkbox to enable Automatic Refresh (old behavior)
     - Add Label 'Searching...' while waiting for server
-    - Change 'Min rating' steps to 100 
+    - Change 'Min rating' steps to 100
+* Fix 'View Replays in Vault' change to 'Online Vault' Tab (#495)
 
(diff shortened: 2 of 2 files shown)


src/chat/chatter.py
```diff
@@ -387,6 +387,7 @@ def viewReplay(self):
     @QtCore.pyqtSlot()
     def viewVaultReplay(self):
         """ see the player replays in the vault """
+        self.lobby.client.replays.setCurrentIndex(2)  # focus on Online Fault
```
duk3luk3, line 387: Is there a reason this isn't grouped with the `self.lobby.client.mainTabs.setCurrentIndex` call?

duk3luk3:
Tested, does what it says.

Before this change: Selecting "View Replays in Vault" in a chat member's context menu switched to whatever Replay Vault Tab was selected last. 
After: It always switches to the "Online Vault" Tab so you can actually see what you wanted to see.

In the future, please put a description like this into PRs.

src/chat/chatter.py
GrotheFAF: the logic behind being first is: 1.get there, 2. do stuff.

duk3luk3: but then `self.lobby.client.mainTabs.setCurrentIndex` should also be part of "get there", shouldn't it?

GrotheFAF: well, if I would have moved the line, then the question would have been why I move the line.
I can't win here.

duk3luk3: Where do you think the line should be?

GrotheFAF: Question of taste, does it matter?

duk3luk3: you said yourself that it should be "1. get there, 2. do stuff".

GrotheFAF: so?

GrotheFAF: again, does it matter?

Wesmania: `self.lobby.client.mainTabs.setCurrentIndex` switches client tab to the replays tab, it's conceptually different than setting up the replay tab. It's fine the way it is, go ahead and merge it.

duk3luk3:
Please rebase and add changelog entry.

PR #497, opened 2016-11-17
fix duration for live games in Replays

GrotheFAF:
Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/replays/replayitem.py
+++ b/src/replays/replayitem.py
@@ -248,7 +248,9 @@ class ReplayItem(QtGui.QTreeWidgetItem):
                     if self.winner is not None:  # FFA in rows: Win ... Lose ....
                         teams += self.FORMATTER_REPLAY_FFA_SPOILED.format(winner=winnerHTML, players=players)
                     else:
-                        if self.teamWin == team:
+                        if "playing" in self.duration:
+                            teamTitle = "Playing"
+                        elif self.teamWin == team:
                             teamTitle = "Win"
                         else:
                             teamTitle = "Lose"
--- a/src/replays/replayitem.py
+++ b/src/replays/replayitem.py
@@ -126,9 +126,9 @@ class ReplayItem(QtGui.QTreeWidgetItem):
         if message['end'] == 4294967295:  # = FFFF FFFF (year 2106) aka still playing
             seconds = time.time()-message['start']
             if seconds > 86400:  # more than 24 hours
-                self.duration = str(int(seconds/86400)) + " days<br />&nbsp;! error !"
+                self.duration = "<font color='darkgrey'>end time<br />&nbsp;missing</font>"
             elif seconds > 7200:  # more than 2 hours
-                self.duration = time.strftime('%H:%M:%S', time.gmtime(seconds)) + "<br />? error ?"
+                self.duration = time.strftime('%H:%M:%S', time.gmtime(seconds)) + "<br />?playing?"
             elif seconds < 300:  # less than 5 minutes
                 self.duration = time.strftime('%H:%M:%S', time.gmtime(seconds)) + "<br />&nbsp;<font color='darkred'>playing</font>"
                 self.live_delay = True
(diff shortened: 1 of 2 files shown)


src/replays/replayitem.py
```diff
@@ -122,7 +122,18 @@ def update(self, message, client):
...
         self.name      = message["name"]
         self.mapname   = message["map"]
-        self.duration  = time.strftime('%H:%M:%S', time.gmtime(message["duration"]))
+        if message['end'] == 4294967295:  # = FFFF FFFF (year 2106) aka still playing
+            seconds = time.time()-message['start']
+            if seconds > 86400:  # more than 24 hours
+                self.duration = str(int(seconds/86400)) + " days<br />&nbsp;! error !"
+            elif seconds > 7200:  # more than 2 hours
+                self.duration = time.strftime('%H:%M:%S', time.gmtime(seconds)) + "<br />? error ?"
+            elif seconds < 300:  # less than 5 minutes (color darkred is used as flag in onlineTreeDoubleClicked)
```
duk3luk3: Don't use that as a flag, define an actual flag.

src/replays/replayitem.py
```diff
@@ -122,7 +122,18 @@ def update(self, message, client):
         
         self.name      = message["name"]
         self.mapname   = message["map"]
-        self.duration  = time.strftime('%H:%M:%S', time.gmtime(message["duration"]))
+        if message['end'] == 4294967295:  # = FFFF FFFF (year 2106) aka still playing
+            seconds = time.time()-message['start']
+            if seconds > 86400:  # more than 24 hours
+                self.duration = str(int(seconds/86400)) + " days<br />&nbsp;! error !"
```
duk3luk3: If you think something is an error, raise an exception / log it / tell the user with a dialog. Don't do something like this.

src/replays/replayitem.py
```diff
@@ -122,7 +122,18 @@ def update(self, message, client):
...
+            elif seconds > 7200:  # more than 2 hours
+                self.duration = time.strftime('%H:%M:%S', time.gmtime(seconds)) + "<br />? error ?"
```
duk3luk3: Same goes for this one obviously.

duk3luk3 (changes requested):
Thanks for the PR, please fix the things I commented on.

src/replays/replayitem.py
GrotheFAF: this is a replay information. no use alarming anyone

src/replays/replayitem.py
GrotheFAF: this is a replay information. no use alarming anyone

src/replays/replayitem.py
GrotheFAF: that is an actual flag, but if you insist.

src/replays/replayitem.py
duk3luk3: if you don't want to alarm people, why are you putting "! error !" there? I agree that it's not a critical problem for the client that should lead to the client shutting down, but putting "! error !" or "? error ?" into the display just sucks.

GrotheFAF: so what do you suggests?

duk3luk3: log a warning, but otherwise ignore it.

GrotheFAF: what would be the purpose of that?

duk3luk3: It will be in the log in case it causes problems somewhere else (say, someone tries to launch the live replay and it doesn't work because it's invalid - we'd see that in the log and could check if it's related)

GrotheFAF: that is beside purpose here. I want to inform the user.
I could reduce it to stating the fact 'no endtime'.

duk3luk3 (approved):

duk3luk3:
Please rebase this and I'll merge it

GrotheFAF:
rebased.

Issue #485, opened 2016-11-14
Let the installer set permissions for ProgramData\FAForever


muellni:
Some people reported to have no write access to `C:\ProgramData\FAForever` and need to run the client as Administrator. The installer MUST fix it, since it introduced #401 .
Could be fixed along with #470 once we figure out how to run admin scripts on installation.
PR #484 tries that...

duk3luk3:
Here is some information about setting permissions: http://msiworld.blogspot.de/2008/11/different-ways-of-giving-permissions-in.html

using any of those options probably means hacking up the bdist_msi module.

GrotheFAF:
I run FAF 'as admin' (shows the mostly empty client), after that I could run FAF normal again. (all .log files are still owned by user)
then I deleted forever.log and repeated it -> a forever.log was created under owner admin and FAF could not start as normal user after that.
So I would guess (as normal user can't take that owner ship) log file creation/rotation should make sure that owner of forever.log is proper inherited.
I used this to see the owner: > dir %ProgramData%\FAForever\logs /q

GrotheFAF:
if it is just about forever.log, there should be something like this in log-rotation after log-rotation and new forever.log creation:
if ctypes.windll.shell32.IsUserAnAdmin():
    os.system("icacls %ProgramData%\FAForever\logs\forever.log /setowner %USERNAME%")
    os.system("icacls %ProgramData%\FAForever\logs\forever.log /reset")
I tested this (with dosbox, not in client code) and it works.

PR #521, opened 2016-12-02
[WIP] Fix programdata write permissions

### LINKED ISSUE #487: ICE test client - FAF cannot be run as administrator error on startup
downloaded the test ICE client build from:

https://ci.appveyor.com/api/buildjobs/gidr2y96896qj061/artifacts/dist%2FForged%20Alliance%20Forever-0.11.60-win32.msi

installed the MSI over top of production client (ie. accepted default install options)
OS is windows 7 pro

after install ran the FAF shortcut from desktop and got below error

image

when close is selected the error below displays:

image

faf log:

[forever.zip](https://github.com/FAForever/client/files/589871/forever.zip)

tried to add full control to the user for the path : C:\ProgramData\FAForever but no change in behavior seen

muellni:
As discussion to fix #487 and #485.

What this PR fixes:
I have two Users A (privileged) and B (unpriviledged).
With User B I did
1. Remove programdata\faforever.
2. Start and stop the client
3. Log into User A
4. Start the client -> fails with write permissions

### DIFF
--- a/src/__main__.py
+++ b/src/__main__.py
@@ -60,12 +60,17 @@ def excepthook(exc_type, exc_value, traceback_object):
     sys.excepthook = excepthook
 
 def AdminUserErrorDialog():
-    box = QtGui.QMessageBox()
-    box.setText("FAF cannot be run as an administrator!<br><br>This probably means you need to fix the file permissions in C:\\ProgramData.<br>Proceed at your own risk.")
-    box.setStandardButtons(QtGui.QMessageBox.Close)
-    box.setIcon(QtGui.QMessageBox.Critical)
-    box.setWindowTitle("FAF privilege error")
-    box.exec_()
+    from config import Settings
+    ignore_admin = Settings.get("client/ignore_admin", False, bool)
+    if not ignore_admin:
+        box = QtGui.QMessageBox()
+        box.setText("FAF should not be run as an administrator!<br><br>This probably means you need to fix the file permissions in C:\\ProgramData.<br>Proceed at your own risk.")
+        box.setStandardButtons(QtGui.QMessageBox.Ignore | QtGui.QMessageBox.Close)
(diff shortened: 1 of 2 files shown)


duk3luk3:
Looks good! This doesn't do anything in linux, right?

It would also be good if the user got more information about what's going on - maybe a link to a wiki page that explains what the problem is in the dialog?

duk3luk3:
Actually, this won't fix the (probably) most common issue: When the client was only started randomly as admin once. You should probably make the check recursive, or do it whenever the client hits a `PermissionDenied` error.

muellni:
I tested to run FAF as admin in between and it still worked as non-priviledged user afterwards.

muellni:
But we can expand the user-test to also test certain files instead of only the Programdata\faforever folder.

muellni:
Can you find some testsubject which still has problems with this build?

muellni:
As proposed by duk3luk3, we should keep the "don't run faf as admin" warning, but make it overrideable
like a "don't show this again" checkbox or a setting in FA Forever.ini

And PermissionDenied exceptions have to be properly handled by running the permission fix.

muellni:
The you're-running-as-admin warning now has an ignore button, which makes the warning disappear for good

duk3luk3:
I'd just like to write down somewhere what exactly this PR is now intended to do:

* Show this warning if FAF is running as admin: `"FAF should not be run as an administrator!<br><br>This probably means you need to fix the file permissions in C:\\ProgramData.<br>Proceed at your own risk."`
* Adds a `check_data_path_permissions` function that checks if `C:\ProgramData\FAForever` is owned by the current user
* Adds a `set_data_path_permissions` function that changes the owner of all files in `C:\ProgramData\FAForever` to the current user
* `set_data_path_permissions` is called when `C:\ProgramData\FAForever` is not owned by the current user or when `faforever.log` cannot be opened for writing.

PR #528, opened 2016-12-16
fix whatsnew

### LINKED ISSUE #524: Use new website
Website has been switched over, 'What's new' tab and other things need to be pointed there now.

muellni:
Fix #524 by installing required SSL dlls and setting the Whatsnew-URL


duk3luk3:
TODO:

* make whatsnew url not hardcoded
* add debug logging?

Will try to do that on the weekend.

PR #529, opened 2016-12-16
fix #523 by checking the WMI service and show error messages accordingly

### LINKED ISSUE #523: uid-check needs the WMI service running on Windows
On Windows the uid-check (by `uid.dll` or `faf-uid.exe`) needs the `Windows Management Instrumentation` service running/usable  to query system information. The client would better check the service with code like
```
try: 
    win32serviceutil.QueryServiceStatus('Winmgmt')
except:
    print "Windows service not installed"
```
to inform the user about a bad system state. Otherwise all the user's system attributes would become "invalid" and he'd be flagged as smurf.
See https://github.com/FAForever/uid/issues/7 .

muellni:
fix #523

### DIFF
--- a/src/util/__init__.py
+++ b/src/util/__init__.py
@@ -4,10 +4,13 @@ import subprocess
 import getpass
 from ctypes import *
 
-from PyQt4.QtGui import QDesktopServices
+from PyQt4.QtGui import QDesktopServices, QMessageBox
 
 from config import Settings
 from PyQt4.QtGui import QDesktopServices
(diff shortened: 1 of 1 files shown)


duk3luk3:
I you rebase this on top of ef44cef03e395617bbaa0e6939c967245382138e I will ff-merge it before pushing it into develop.

muellni:
I rebased it on 12abf2393810be9daf372003ce9663e02e105e67 hope that's what you wanted...

PR #530, opened 2016-12-16
#524 Use the new "News page"

micheljung:
And remove the _World Stopping_ "Download Client" button that confuses people


duk3luk3:
You need to cherry-pick this commit: d71c3ff3cc55b8ba035a87b5ad733610bd3d22a4

you'll also need another one.

duk3luk3:
Now you need to fixup runtests.py to set all the api's to version 2, like in the new runtests.py: https://github.com/FAForever/client/blob/develop/runtests.py

EDIT: Actually, the old runtests.py doesn't do any api setting at all... so no idea.

duk3luk3:
Merged into `release/0.11.64`

Wesmania:
The PR branch has some no longer used hotfix commits on top - I'll take the liberty to manually delete it.

PR #547, opened 2016-12-26
catch exceptions in check_data_path_permissions()

muellni:
Try to fix http://forums.faforever.com/viewtopic.php?f=3&t=13728


muellni:
- [x] better exception logging https://docs.python.org/2/library/traceback.html 
- [x] test it by raising some dummy exception

PR #551, opened 2016-12-29
Make __main__.py a runnable-anywhere executable scipt

### LINKED ISSUE #550: Modify src/__main__.py so it can be run as linux executable
Making \_\_main\_\_.py a script (via a shebang) that's runnable anywhere greatly simplifies installation on Linux distros. The script can be automagically picked up by the build system (as it does on Gentoo, from which this change comes from) or can be simply copied anywhere and run directly.

Wesmania:
This closes #550 and makes installation of FAF on linux distros
much easier - now you can just copy over __main__ to $PATH rather
than make brittle scripts with hardcoded path to site-packages.

Thank you for contributing to this project! Please set a good title for your PR and  replace this line with a (preferrably multi-line) description of your PR.

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,7 +36,7 @@ install:
   - "%PYTHON%\\Scripts\\pip install cffi"
   - "%PYTHON%\\Scripts\\pip install cx_Freeze"
   - "%PYTHON%\\Scripts\\pip install -r requirements.txt --trusted-host content.faforever.com"
-  - ps: "$env:APPVEYOR_BUILD_VERSION = (& C:\\Miniconda\\python .\\src\\config\\version.py .) + \"+\" + $env:APPVEYOR_BUILD_NUMBER"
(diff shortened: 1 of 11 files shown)


duk3luk3:
> but really the best way to fix it would be to convert our imports to relative form.

Is there anything stopping us from doing that? Which imports specifically need to be fixed?

Wesmania:
@duk3luk3 Pretty much everywhere where we import any local packages - thing like 'import util' and 'import config'. Changing these to 'from {.,..} import foo' would be one way of fixing it.
The logic in this patch is still useful for inclusion into a small script that only sets up path and calls a main method from the package (the way youtube-dl does it, for example) - it'd just require extracting all the logic from \_\_main\_\_.py into a function in \_\_init\_\_.

Wesmania:
Now that I'm trying to do some relative imports, it seems much harder than I thought. I'd love to hear an opinion of someone more versed in Python, but I think adding the module to python path might be the more 'right' solution after all.

duk3luk3:
Well it should just be installed as a proper module. I think we should put that on the list of things to do when we finally migrate to python 3.

PR #553, opened 2016-12-30
552 res dir cleanup

### LINKED ISSUE #552: Clean up working and resources directory finding logic
There is some code in \_\_main\_\_.py that sets up client's current directory. This directory is only used to store the RELEASE-VERSION file. If we move this file to 'res', we won't need to change directory, only to find the resources directory. Logic for this can be in turn extracted to a module.

Wesmania:
This fixes #552, cleaning up all logic regarding locating the resources dir / changing directories.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,7 +36,7 @@ install:
   - "%PYTHON%\\Scripts\\pip install cffi"
   - "%PYTHON%\\Scripts\\pip install cx_Freeze"
   - "%PYTHON%\\Scripts\\pip install -r requirements.txt --trusted-host content.faforever.com"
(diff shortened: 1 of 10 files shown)


duk3luk3:
@muellni What do you think about this?

@Wesmania please add a changelog entry.

Wesmania:
@duk3luk3 Where exactly should I add it in changelog.md? There's no tag yet for any future version, should I add something?

duk3luk3:
Just add it to 0.12 :)

Wesmania:
It turns out there's some convoluted logic in version.py that can't import fafpath, since it's run directly by build scripts. I reworked it to fit with the new res directory finding logic.

Issue #548, opened 2016-12-28
wrong faction from ladder


shalkya:
on 12.0 pre 2 (last one)

http://imgur.com/BiYzGB6

content.faforever.com/faf/vault/replay_vault/replay.php?id=5585743

GrotheFAF:
not clear what the actual conditions were, user claims he didn't chose UEF but got UEF.

duk3luk3:
I fixed *something*. Please take a look at this @GrotheFAF, what was the problem you said you saw?

Wesmania:
Is this still an issue?

shalkya:
I didn't hear any bug similar to this one occurring again recently

PR #554, opened 2016-12-30
Fix/ranked search select

duk3luk3:
Potential fix for #548.

The race selection was buggy and did not register the first click
properly.
The stringly typing was removed and that fixed the issue, somehow.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


GrotheFAF:
look complicated, I just added an init line in _gamewidget.py L55:
 Settings.set("play/subFactions", self.sub_factions)  # needs one write before self.sub_factions... work

PR #565, opened 2017-01-06
Handle case when user is not logged in in Players

### LINKED ISSUE #564: Local replays marked as "broken" when offline
You can't watch local replays from FAF when offline, since the replays show up as "broken". This is because of a catch-all catching an exception elsewhere in the code, which is thankfully easily fixable.

Wesmania:
Players class handles mapping of various IDs to players and their
relations to the logged in player ('me'). If we're not logged in, 'me'
stays null. We should account for that.

This fixes #564.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -30,6 +30,7 @@
 * Clean up 'res' directory finding (#552)
 * Improve ladder race selection (#554)
 * Make selected mods persistent (#560)
+* Disable modvault feedback (#568)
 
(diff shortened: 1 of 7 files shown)


duk3luk3:
👍 Will review when I can

duk3luk3:
Please rebase and add changelog!

duk3luk3:
Merged into develop (f602747)

PR #568, opened 2017-01-07
Disable Modvault feedback

### LINKED ISSUE #559: like and comment mods (mod vault)
http://forums.faforever.com/viewtopic.php?f=40&t=13708

comment doesn't work
like doesn't work and make you disconnect.

why ? 
server doesn't support old method of interacting with mod vault anymore
it tells the client that it's an invalid operation
and invalid operations cause the client to disconnect

we need to either make server and mod vault interact again; or we remove like and comment from the feature.

duk3luk3:
Disables linking / commenting / bugreporting on mods.

This is a temporary measure because the new server doesn't support the
requests the client sends for that functionality.
In the long term, these features will use the API.

Fixes #559

- [ ] Code has tests
- [X] Final commit includes "Fixes #issue" in commit message

- [X] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -30,6 +30,7 @@
 * Clean up 'res' directory finding (#552)
 * Improve ladder race selection (#554)
 * Make selected mods persistent (#560)
+* Disable modvault feedback (#568)
 
 0.11.64
 =======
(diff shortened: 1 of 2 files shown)


micheljung:
What requests did the client send? I thought it's all handled in PHP.

duk3luk3:
No, it used to be implemented in the lobbyserver. Then it was dropped, resulting in any of the mod-related commands leading to an "INVALID" response and a kick from the server.

micheljung:
Aaah that one x) alright.

PR #572, opened 2017-01-07
571 paths cleanup

### LINKED ISSUE #571: Make running FAF independent from running directory
It'd be convenient for FAF to not need to be run while in a specific directory to work properly. We can deduce all the directories we need from things like sys.executable and __file__.

Wesmania:
This PR does some more cleanup regarding figuring out paths to FAF files, guessing if we're running from source and using the 'lib' directory. In particular it fixes #571.


Wesmania:
MSI tested by EcoNoob, everything seems to work fine.

PR #586, opened 2017-01-14
0.12.0 Release candidate

duk3luk3:
This PR just adds a list of contributors. I just added the ones I know about. Do any others need to be added?

Does anything else need to be done for the final 0.12.0 release? Please note it here!

EDIT: Please also submit testing reports here.


duk3luk3:
If you only want to say "I used the new client for several days and found no problems" please just thumbs-up this comment.

PR #587, opened 2017-01-15
Make the Reload Stylesheet Menu Item reload all Stylesheets

### LINKED ISSUE #584: make theme stylesheet reload better-er
**I experienced...** a great feature that lets you reload the stylesheet of a current theme without restarting the client... but it only reloads one style sheet and not the multiple a theme might have!

**how to reproduce** get a theme with mutliple stylesheets and reload the stylesheet with the menu option. any changes made after client restart in the other stylesheets wont reload.

**I expected...** all the default stylesheets to reload

**Where the bug is / What should be done**
there is a great commit that makes it work here: https://github.com/madformuse/client/commit/1ce2fd66dee1e5d5c8b4af0112d8475898a0044e
this only applies to the tourney tab, so simply going through the _res folder and adding all the stylesheets in the above manner would work just fine.

yorick-ne:
Makes the Reload Stylesheet Menu Item reload all Stylesheets. 
The Qt Objects that get a nondefault stylesheet are being remembered by encapsulating setStyleSheet.

Fixes #584

Initial PR:

- [ ] Code has tests

- [x] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,11 +11,13 @@
 * Fix typo preventing rehost feature from working (#630 / #632)
 * Remove needless abstract base class for client (#629)
 * Fall back to APPDATA if Documents folder cannot be created (#619, #626)
+* Make all (themed) stylesheets reloadable (#584)
 
 Contributors:
   - Duke
   - speed2
   - Wesmania
+  - yorick
 
 0.12.2
 ======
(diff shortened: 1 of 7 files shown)


duk3luk3:
Very clever, I like it.

duk3luk3:
Does this work? Has it been tested?

PR #499, opened 2016-11-17
feature - single line the layout of modvault

GrotheFAF:
put all textboxes and buttons in a single line
https://github.com/FAForever/client/pull/499/commits/c05349f7d5f2e4822960861eb484206a5ece7e95 remove empty/obsolete layout elements

- [x] PR branch should be named `issuenum`-`fix`/`feature`/`cleanup`-`description`
- [x] Code is split into logical commits
- [ ] Code has tests
- [ ] Final commit includes "Fixes #issue" in commit message

When all builds pass and a maintainer is happy with your PR, the "ready" label will be applied. Please complete these tasks then:

- [x] Rebase onto develop
- [ ] Add changelog entry
- [ ] Remove this entire template section

PR #591, opened 2017-01-25
fix - observer in hosted games #535

### LINKED ISSUE #455: make 'gone offline' Hosts visible in hosted Games in Play
Currently you can't see if a host went offline. (only if you check if host is in chat online)
So more user join and wait  (and evt. leave) -> frutrating waste-o-time.
One could mark or hide the game (so no one else joins)
Automatically pm the joined user about the situation.

First step could be give the Player another color (black/red?) that might not be effectiv for randon color user, so be explizit: '[Player] (offline)'.

### LINKED ISSUE #535: Player VS VS
**I experienced...** hovering over a game in the games tab with 3 players showed this:
untitled

**how to reproduce** not sure

**I expected...** not a second VS and actually a list of the other players in the game.

**Where the bug is / What should be done** if you have a suggestion for fixing the issue, add it here.

GrotheFAF:
for hosted games in Play:
- Reintroduce observer in tooltip - fixes #535 
- Sort out code only needed while hosting (state = 'open')
- Add extra-teamup info to hosted game
- Make host offline more visible - fixes #455 (kinda)

some variable renaming

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/games/gameitem.py
+++ b/src/games/gameitem.py
@@ -313,44 +313,44 @@ class GameItem(QtGui.QListWidgetItem):
 
     def editTooltip(self, teams, observers):
         
-        teamlist = []
+        teams_list = []
 
         i = 0
         for team in teams:
             
             i += 1
 
-            teamplayer = ["<td><table>"]
+            players_list = ["<td><table>"]
             for player in team:
 
                 if player == client.instance.me:
-                    playerStr = "<b><i>%s</b></i>" % player.login
+                    player_str = "<b><i>%s</b></i>" % player.login
                 else:
-                    playerStr = player.login
+                    player_str = player.login
 
                 if player.rating_deviation < 200:
-                    playerStr += " (%s)" % str(player.rating_estimate())
+                    player_str += " (%s)" % str(player.rating_estimate())
 
                 country = os.path.join(util.COMMON_DIR, "chat/countries/%s.png" % (player.country or '').lower())
 
                 if i == 1:
-                    player_tr = "<tr><td><img src='%s'></td>" \
-                                    "<td align='left' valign='middle' width='135'>%s</td></tr>" % (country, playerStr)
+                    player_html = "<tr><td><img src='%s'></td><td align='left' " \
+                                  "valign='middle' width='135'>%s</td></tr>" % (country, player_str)
                 elif i == self.nTeams:
-                    player_tr = "<tr><td align='right' valign='middle' width='135'>%s</td>" \
-                                    "<td><img src='%s'></td></tr>" % (playerStr, country)
+                    player_html = "<tr><td align='right' valign='middle' width='135'>%s</td>" \
+                                    "<td><img src='%s'></td></tr>" % (player_str, country)
                 else:
-                    player_tr = "<tr><td><img src='%s'></td>" \
-                                    "<td align='center' valign='middle' width='135'>%s</td></tr>" % (country, playerStr)
+                    player_html = "<tr><td><img src='%s'></td><td align='center' " \
+                                  "valign='middle' width='135'>%s</td></tr>" % (country, player_str)
 
-                teamplayer.append(player_tr)
+                players_list.append(player_html)
 
-            teamplayer.append("</table></td>")
-            members = "".join(teamplayer)
+            players_list.append("</table></td>")
+            team_html = "".join(players_list)
 
-            teamlist.append(members)
+            teams_list.append(team_html)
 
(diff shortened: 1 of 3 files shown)


duk3luk3:
How do I test this?

GrotheFAF:
passiv: by inspecting hosted games.
activ: by hosting games in different players/team/observer variations.

duk3luk3:
You need to tell me what exactly to do to reproduce the problem this fixes.

GrotheFAF:
Read the Issue #535.
You can reproduce the observer case, which this fixes.
(but not the user-offline case, that's a server thing.)

src/games/gameitem.py
```diff
@@ -169,26 +169,32 @@ def update(self, message, old_client=None):
...
-        num_players = message.get('num_players', 0)
-        self.slots = message.get('max_players', 12)
-        
-        oldstate = self.state
+        if 'host_id' in message:
+            self.hostid = message['host_id']
+        elif self.hostid == -1:  # fresh game
+            self.hostid = client.instance.players.getID(self.host)
+        elif self.hostid != client.instance.players.getID(self.host):  # somethings funny (offline)
+            self.hostid = client.instance.players.getID(self.host)
```
Wesmania, line 187: I really don't like this defensive programming, what's the purpose of doing exactly the same thing in the 'else' branch?

Wesmania:
This PR would really be more readable if it was split into more commits, right now it's code reordering, variable renames and meaningful changes all mixed together.

src/games/gameitem.py
GrotheFAF: These 2 are very different cases, so I want to keep them seperate, even with currently the same 'solution'. 
In the current client in theory the host can't change - so the only cases for the 2nd is: host is shown offline .. which currently happens.

GrotheFAF:
it is split into 2 commits:
1. code reordering, meaningful changes (and ok 4 renames which are easy to spot)
2. variable renames 
let's move on shall we?

GrotheFAF:
so no chance of for this? let me know, I remove it. 
no use dragging this around from rebase to rebase.

duk3luk3:
You still haven't explained what the bug actually is and how your changes fix it. I honestly don't understand it.

GrotheFAF:
The explaination is in the issues. You want it here in the PR too?

Wesmania:
To be honest I'd love to see all the parts of building that table factored off into several classes or functions or anything of the sort rather than complex fixes. Looking at all these HTML strings being concatenated makes my eyes bleed.

GrotheFAF:
so no chance...

duk3luk3:
> The explaination is in the issues.

No, it's not. I asked for a better explanation in the issues and didn't get one. In #455 both Wesmania and me asked for a better explanation and didn't get any useful response from you.

GrotheFAF:
the last comment there was an explantation from me, if you weren't satisfied with that you could have reacted, but you didn't (both) ... so don't shit me.

duk3luk3:
I'm not shitting you. You're not the most important thing in my life, you're not the most important contributor to the FAF client, and you continue to make bad issue reports and bad PRs.

So I don't make dealing with you my top priority.

We've had the same discussion a few months ago, with the same results. You can draw your own conclusions from that.

When you make good reports and good PRs I merge them, like I've done, multiple times.

When you don't, I ask you for clarification or improvements, like I've done, multiple times.

Your comments on #455 are full of speculation about server bugs or connection issues. Instead of speculating and making a PR that solves a problem that nobody knows really exists with workarounds that look weird and don't have any explanation, you should have investigated what exactly causes "user shows up as irc user", and if there are really issues with the server.

That's just not acceptable, and you shouldn't need prodding and pushing to do better.

You're welcome to contribute to this project, and I'm very grateful for the good contributions you've made in the past. But you don't have a right to having your PRs merged, and your attitude sucks.

GrotheFAF:
I think your attitude sucks. I would call it lazy and abuse of power.

for example to call #499 a bad PR is just bad attitude.
to #455. I investigated that - put a lot of time in it (asked for help server side), but as I have no real feel for networking, no conclusive result, so I tried to work around it, did what I could. you call that 'not acceptable'... well. That is some bad attitude. makes me fucking angry.

If I wouldn't have my 'attitude' and fought for the client against all admins, we wouldn't be here working on the client - remember? maybe not, was before you were on slack ...and you wouldn't had that job, with all that power to decide what is good and what is bad.

PR #621, opened 2017-02-18
620 factor out lobby connection

Wesmania:
Move lobby-server-talking code outside _clientwindow.py as much as possible. There's still a conflation of client state and server connection state (client.state), but we won't touch if for now to avoid breaking it.

- [x] Code has tests

- [x] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/src/client/connection.py
+++ b/src/client/connection.py
@@ -24,6 +24,8 @@ class ServerConnection(QtCore.QObject):
 
     # These signals are emitted when the client is connected or disconnected from FAF
     state_changed = QtCore.pyqtSignal(object)
+    connected = QtCore.pyqtSignal()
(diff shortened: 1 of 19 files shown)


src/client/connection.py
```diff
@@ -31,6 +31,31 @@ def __init__(self, connection):
...
+        self._keepalive_timer.timeout.connect(self._ping_connection)
+        self.keepalive_interval = 10 * 1000
+
+    @property
+    def keepalive(self):
+        return self._keepalive
+
+    @keepalive.setter
+    def keepalive(self, value):
+        self._keepalive = self.value
```
muellni: typo: `self._keepalive = value`

PR #639, opened 2017-03-02
add subdirectory support to Local Archive

GrotheFAF:
add a bucket for subfolders in local replays folder with change-folders function

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -407,7 +407,7 @@ class ReplaysWidget(BaseClass, FormClass):
                 item = QtGui.QTreeWidgetItem()
                 item.setText(1, infile)
                 item.filename = os.path.join(replay_dir, infile)
-                item.setText(3, "(" + str(len(os.walk(item.filename).next()[2])) + " files)")  # takes time on first run
+                item.setText(3, "(" + str(len(os.listdir(item.filename))) + " files)")  # takes time on first run
                 item.setIcon(0, util.icon("replays/bucket.png"))
                 item.setTextColor(0, QtGui.QColor(client.instance.getColor("default")))
 
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -416,18 +416,18 @@ class ReplaysWidget(BaseClass, FormClass):
         if len(cache_add) > 10 or len(cache) - len(cache_hit) > 10:
             self.saveLocalCache(cache_hit, cache_add)
         # Now, create a top level TreeWidgetItem for every bucket, and put the bucket's contents into them
-        for bucket in buckets.keys():
+        for bucket_key in buckets.keys():
             bucket_item = QtGui.QTreeWidgetItem()
             
-            if bucket == "broken":
+            if bucket_key == "broken":
                 bucket_item.setTextColor(0, QtGui.QColor("red"))  # FIXME: Needs to come from theme
                 bucket_item.setText(1, "(not watchable)")
                 bucket_item.setTextColor(1, QtGui.QColor(client.instance.getColor("default")))
-            elif bucket == "incomplete":
+            elif bucket_key == "incomplete":
                 bucket_item.setTextColor(0, QtGui.QColor("yellow"))  # FIXME: Needs to come from theme
                 bucket_item.setText(1, "(watchable)")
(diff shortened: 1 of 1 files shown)


src/replays/_replayswidget.py
```diff
@@ -322,7 +347,9 @@ def updatemyTree(self):
 
                     # Parse replayinfo into data
                     if item.info.get('complete', False):
-                        t = time.localtime(item.info.get('launched_at', item.info.get('game_time', time.time())))
+                        # 'game_time' is 'launched_at' for older replays (if that fails maybe 'game_end' is there)
```
Wesmania: If we're tinkering with the 't' and folders already, maybe we can put replays without that info into a separate bucket?

GrotheFAF: without what info and for what purpose?
([partly] broken replays are already treated in seperate buckets)

Wesmania: I once downloaded a replay from forums (4856559) that when put into the replays folder, always shows on the top. It seems to be parsed properly, but this part of the code makes it always appear on top, not in the 'incomplete replays' bucket. Maybe moving it into 'incomplete' bucket is a good idea if we can't find the start date.

GrotheFAF: yes, using current-date as 'last resort' for game-date-time is confusing (because it's 'fake-data').
So an extra bucket 'data missing' or 'date missing' might be right here.

Wesmania:
That replay I was talking about (the one you fixed with the third commit) disappeared entirely from the list now. I think it doesn't get added to any of the buckets.

Wesmania:
TBH I also don't quite understand what the commit precisely does (code-wise), that's probably why the PR is still held up. More comprehensive commit names and descriptions would help a lot.

src/replays/_replayswidget.py
```diff
@@ -370,37 +400,63 @@ def updatemyTree(self):
...
                 bucket.append(item)
 
+            elif os.path.isdir(os.path.join(replay_dir, infile)):  # we found a directory ...
+
+                bucket = buckets.setdefault("directories", [])
+
+                item = QtGui.QTreeWidgetItem()
+                item.setText(1, infile)
+                item.filename = os.path.join(replay_dir, infile)
+                item.setText(3, "(" + str(len(os.walk(item.filename).next()[2])) + " files)")  # takes time on first run
```
GrotheFAF: Is there a better/faster way to count files in a directory?

GrotheFAF:
You don't have to 'think'/guess, you could check the numbers (of your replays/files).
It's about directories, read my comments in the code or ask about what you do not understand.
(And maybe, you know better/faster way to count files in a directory? (line 410))

Wesmania:
I don't think the OS itself keeps file number info, so if we want speed, we'll probably have to drop showing file number.
Also I insist on renaming / adding descriptions to some of these commits, they're pretty large and putting them into history the way they are would be pretty confusing IMO.

GrotheFAF:
This is straight forward code, nothing complicated, so explain what would be 'pretty confusing', be specific, I insist.

Wesmania:
Things like 'add subdirectory support to Local Archive details' don't tell anything about what was changed in the implementation. The first commit itself doesn't have a description how this subdirectory support works (does it replace current date buckets? What directories are shown? Are replays in subdirs also grouped into buckets? Does every bucket potentially have subdirs with replays from that date?)

If the two 'details' commits are just fixups, you could probably just squash them together with the first commit and force push the rebased branch.

GrotheFAF:
Oh, that basic, then I should probably start with an issue.

I thought about the last part, but when I tried something like that (changing commits) I got only more commits (aka more mess). So I guess I close this PR and start from scratch.

Wesmania:
There's no need for that. You can use `git rebase -i develop` to rebase / squash / edit your commits the usual way, then just `git push -f` to push the changes into the PR branch. Github picks up on it and updates the PR just fine. Feel free to reopen and just do that.

GrotheFAF:
As I wrote, I don't know how to squash or edit commits, when I tried it only got messier, meaning more commits, not less.

GrotheFAF:
That rebase/squash and push -f  worked fine .. 
But I can't reopen the PR now. probably should have reopen first ...
seems the case: https://github.com/isaacs/github/issues/361

PR #649, opened 2017-03-05
Don't log null byte in replayserver.py

### LINKED ISSUE #648: Strange symbol in forever.log
When trying to copy paste the whole log I've noticed  that just part of it got pasted even thought I hit Ctrl A to select all. So when I went to the last line I found this there:
https://puu.sh/uuXaw/9206e69be6.png

Not sure how severe that is if all. The replay was a FAF Beta mod where I did some testing for Icecreamer.
When searching replay vault it is there but missing end time
https://puu.sh/uuXgk/d26a4f9c1b.png

Wesmania:
Fixes #648.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/fa/replayserver.py
+++ b/src/fa/replayserver.py
@@ -74,7 +74,7 @@ class ReplayRecorder(QtCore.QObject):
             #This prefix means "P"osting replay in the livereplay protocol of FA, this needs to be stripped from the local file            
             if data.startsWith("P/"):
                 rest = data.indexOf("\x00") + 1
-                self.__logger.info("Stripping prefix '" + str(data.left(rest)) + "' from replay.")
+                self.__logger.info("Stripping prefix '" + str(data.left(rest - 1)) + "' from replay.")
                 self.replayData.append(data.right(data.size() - rest))
             else:
                 self.replayData.append(data)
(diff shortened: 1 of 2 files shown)


src/fa/replayserver.py
```diff
@@ -74,7 +74,7 @@ def readDatas(self):
             #This prefix means "P"osting replay in the livereplay protocol of FA, this needs to be stripped from the local file            
             if data.startsWith("P/"):
                 rest = data.indexOf("\x00") + 1
-                self.__logger.info("Stripping prefix '" + str(data.left(rest)) + "' from replay.")
+                self.__logger.info("Stripping prefix '" + str(data.left(rest) - 1) + "' from replay.")
```
duk3luk3: I think you mean `data.left(rest - 1)`

duk3luk3: Also you could just change this all into

```python
prefix, tail = data.split("\x00", 1)
self.__logger.info("Stripping prefix '" + str(prefix) + "' from replay.")
self.replayData.append(tail)
```

Wesmania: We can't, unfortunately - it's a QByteArray, not a string.

duk3luk3: `prefix, tail = str(data).split("\x00", 1)` then?

Wesmania: It's later appended to another QByteArray - I don't know if we want to convert it to Python string and back just for some nice syntax.

Wesmania: Arguably it only happens once per replay, but that conversion still feels less tidy to me than just working with what QByteArray has.

PR #653, opened 2017-03-10
652 make login wizard a widget

### LINKED ISSUE #652: Clean up login window widget
Right now the login window widget is a subclass of QWizard, which causes way too much ugly layouting code and limits us when it comes to fancy layouts and theming. We can do much better if we just make it a QDialog with its own layout.

Wesmania:
This replaces the login window based on a QWizard with a proper QDialog with layout, emulating the old window as closely as possible. Fixes #652.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


res/client/login.css
```diff
@@ -0,0 +1,4 @@
+QWidget#header
+{
+    background-color: rgba(255,255,255,255);
```
GrotheFAF, line 3: An RGBA color value is specified with: rgba(red, green, blue, alpha). The alpha parameter is a number between 0.0 (fully transparent) and 1.0 (fully opaque).

Wesmania: According to [Qt docs](http://doc.qt.io/qt-4.8/stylesheet-reference.html): "The rgb() or rgba() syntax can be used with integer values between 0 and 255, or with percentages. "

PR #661, opened 2017-03-15
beautify code on files in fa subdir

GrotheFAF:
mostly blank lines and comment corrections
 lowercase locally used variables and functions
 (no external references were harmed during the making of this commit)

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/fa/replay.py
+++ b/src/fa/replay.py
@@ -3,9 +3,8 @@ import os
 from PyQt4 import QtCore, QtGui
 import fa
 from fa.check import check
-from fa.replayparser import replayParser
+from fa.replayparser import ReplayParser
 import util
-import mods
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


src/fa/replay.py
```diff
@@ -14,26 +13,28 @@
...
+    """
     logger.info("fa.exe.replay(" + str(source) + ", detach = " + str(detach))
 
     if fa.instance.available():
         version = None
         featured_mod_versions = None
         arg_string = None
         replay_id = None
+        mod = None
+        map = None
```
Wesmania: 'map' is a Python keyword, we should avoid that.

Wesmania:
I like this. Do we have coding standards for underscores / camelcase? It might be a good idea to set some up.

src/fa/replay.py
GrotheFAF: oh, missed that  .. commit comin'.

GrotheFAF:
This is after python rules in PyCharm:
camels and 2 blank lines for classes, everthing else is lowercase and max. 1 blank line.
(Qt and other Libs are still Pascal-ish and PyCharm is fine with it.)

GrotheFAF:
and this will not be processed too.

PR #500, opened 2016-11-17
feature - change to flat style

GrotheFAF:
affects dropdownboxes and pushbuttons:
Tab Play: Sort Games, 
Tab Replay: Search: Mod, Min. Rating, Refresh Recent List, Search
Tab Vaults: Mods: everything  on top (except searchinput)
Host Game: Map
FA bottom: Ladder offer.

- [x] PR branch should be named `issuenum`-`fix`/`feature`/`cleanup`-`description`
- [x] Code is split into logical commits
- [ ] Code has tests
- [ ] Final commit includes "Fixes #issue" in commit message

When all builds pass and a maintainer is happy with your PR, the "ready" label will be applied. Please complete these tasks then:

- [x] Rebase onto develop
- [ ] Add changelog entry
- [ ] Remove this entire template section

PR #668, opened 2017-03-17
feature - add subdirectory support to Local Archive

GrotheFAF:
+ Sub directories in the local replays folder will be listed each as directory item.
Doubleclick one of the subfolders will show the content of that subfolder as current replay folder.
If the current folder is a sub folder (not util.REPLAY_DIR) a 'directory up' item '..' is added. A doubleclick will show the parent folder again as current replay folder.
+ Folders will not be reread on Tab reentry if nothing has changed (using folders mtime).
+ Replays with missing Dates get an seperate Bucket 'date missing'. (untested, I have none of those ;-)

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


Wesmania:
I tried it out, I like the idea. Moving around folders is a little bit confusing, since the ".." element is in a submenu. It'd be great if we just had a separate icon for directories - we could just put all the directories above all the other buckets, with optional '..' as the first position (like in file managers, we see '..', then  directories, then files).

GrotheFAF:
I changed it. 
(but you won't like my sorting trick ;-)

Wesmania:
That's one way to do that. I like the flat model. I think this is pull ready, maybe with a different dir icon. After ICE™ probably.

GrotheFAF:
If you really like flat, why not process #500 ?-)
The folder icons are from XP Explorer, I don't know what ICE™ is. hint please.

Wesmania:
The ICE adapter, major PRs will go in after it's integrated.

GrotheFAF:
Oh, that ICE. Tested and ready?
(but that you consider this PR 'major', makes me laugh. It's tiny.)
I have so much stuff in the pipeline to PR .. but my curent PRs are already stuck .. so WTF?

Wesmania:
Talk to @duk3luk3 about it. I have some PRs queued too, I just have some additional branches on top of those. It does make sense IMO given that ICE is a big feature to add.

duk3luk3:
* What's the license of these icons? We can't pull things that aren't under a compatible free license into a GPL project.
* Is this something users want?

GrotheFAF:
exchanged the icons with GPL once .. and in black.

GrotheFAF:
and what about this, delete it?

PR #677, opened 2017-03-22
Improve logging

duk3luk3:
* Add 'client\log\stderr' setting to log to stderr even if not in
development env
* Make server message logspam level lower than debug

- [ ] Code has tests


duk3luk3:
cx_freeze is bullshit so this isn't really possible to do without bullshit.

PR #688, opened 2017-04-05
Replace game quality in the play tab with average rating

### LINKED ISSUE #687: Replace "game quality" for a game in play tab with something else
A short discussion [here](http://forums.faforever.com/viewtopic.php?f=2&t=14329) suggests nobody really uses game quality indicator for a game in the play tab, and it's not accurate for game that isn't full anyway. We could probably replace it with something more useful.

Wesmania:
Cursory talk with players suggested that hardly anyone uses this number
in the play tab. Average rating might be of more use.

Fixes #687.

- [ ] Code has tests

- [ ] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/res/games/formatters/faf.qthtml
+++ b/res/games/formatters/faf.qthtml
@@ -9,8 +9,8 @@
 <font size="-1">by</font>&nbsp;<font color="{color}"><b>{host}</b></font></font>
 </td>
 <td align="right">
-<b><font color="silver" size="-1">Game Quality</font><br/>
-<font color="silver" size="+2">{gamequality}</font></b>
+<b><font color="silver" size="-1">Avg. Rating</font><br/>
+<font color="silver" size="+2">{avgrating}</font></b>
 </td>
 </tr>
 </tbody>
--- a/res/games/formatters/faf.qthtml
+++ b/res/games/formatters/faf.qthtml
@@ -10,7 +10,7 @@
 </td>
 <td align="right">
 <b><font color="silver" size="-1">Avg. Rating</font><br/>
-<font color="silver" size="+2">{avgrating}</font></b>
+<font color="silver" size="+1">{avgrating}</font></b>
 </td>
 </tr>
 </tbody>
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 * Clean up theme-specific util code (#663, #664)
 * Cleanup receiving game info (#708, #709)
 * Add Friend/Foe feature for IRC users (#609, #657)
+* Replace game quality in the play tab (NOT in the lobby) by average rating (#687, #688)
 
 Contributors:
  - Wesmania
--- a/res/games/formatters/mod.qthtml
+++ b/res/games/formatters/mod.qthtml
@@ -10,7 +10,7 @@
 </td>
 <td align="right">
 <b><font color="silver" size="-1">Avg. Rating</font><br/>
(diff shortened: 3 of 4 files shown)


GrotheFAF:
Agreed, Gamequality seems not to make much sense to joiners, because the final quality will be determent during (hopefully) ballancing, when the game is full any way.
(what about the 'hardly' minority? Maybe the get something out of it that we are missing.)

Wesmania:
That 'hardly' is an estimation, at least no one in the forum thread said they're using it. I gave the issue 'waiting for feedback' tag just in case. I'd like to wait for more devs to give their input before we change it.

GrotheFAF:
ok. 
I would suggest to take the size of avgrating down a nodge from size="+2" to "+1".
and maybe rounding the avgrating to 100?

and maybe we could loose self.nTeams and replace the last remaining use with len(teams).

Wesmania:
I tried out the smaller font, it does seem more aesthetic. Rounding the rating's a good idea IMO, but it's highly visible, so I'll wait for input from others for now.

As for self.nTeams, I think there's some more stuff that can be fixed here. I don't see self.teams used either, and update could maybe take keyword args instead of a dict. In general we could use the philosophy that GameItem does not store state that is updated (teams / players / etc.), but only creates proper URLs / UI from what we give it. It all deserves a fuller refactor IMO.

If it was purely behind-the-scenes change, or new functionality, I'd be happy to try and sneak in refactor commits before the usual ones - but this one changes existing functionality. I'd rather put a potentially thing breaking refactor in a separate PR, and keep this one just for discussion about changing the visible stuff - to make reviewing and discussion simpler.

GrotheFAF:
not self.teams (but hat can go too), just (lokal) teams (-> self.nTeams = len(teams) and used once in edit_tooltip).
but probably better a seperate PR ... not to overload the Duke.

Wesmania:
Yeah, I know that - if we're refactoring some stuff anyway, we could as well refactor it all in a nice chain of commits.

And trust me, there's no need deriding duke - even small fixes can pile up into incomprehensible mess quickly, which then has trouble getting reviewed (Learned that by sending patches to LKML).

GrotheFAF:
well, what else can the powerless do?

duk3luk3:
I don't like cutting out functionality that is working perfectly. Could you instead make this configurable so user can select if they want to see quality / avg. rating / both?

Wesmania:
I realize that - point is, it's only nominally working perfectly. 90% of the time it gives no result at all or no meaningful result. I feel that literally nobody is using it because of that - making it configurable would complicate the ui and greatly complicate the code (you'd need to fiddle around a lot with qthtml) with no benefit at all.

duk3luk3:
OK, I see. I'm fine with cutting it out then.

res/games/formatters/faf.qthtml
```diff
@@ -9,8 +9,8 @@
 <font size="-1">by</font>&nbsp;<font color="{color}"><b>{host}</b></font></font>
 </td>
 <td align="right">
-<b><font color="silver" size="-1">Game Quality</font><br/>
-<font color="silver" size="+2">{gamequality}</font></b>
+<b><font color="silver" size="-1">Avg. Rating</font><br/>
+<font color="silver" size="+1">{avgrating}</font></b>
```
GrotheFAF, line 13: I currently tent again more to size="+2" than "+1" .. but it's a taste thingy.
(same for mod.qthtml)

Issue #704, opened 2017-04-22
Some user in chat userlist are shown in zombie games


GrotheFAF:
Some users in chat are shown 'in game' (crossed sword), but the games are actually older zombie games of these users.
A workaround (until this is fixed server side) would be, not to process/ignore the zombie games (>12 hours) when getting the games from server on client startup.

Wesmania:
We've talked a while about it already, I think? We even located the cause of the bug on the server, and fixing it there would be the right thing to do. It'd be easier to deploy the fix on the server rather than fix it with a client release that will have to be reverted once the bug is fixed.

I know that the fix won't make it to the server before ICE, since it touches the server code quite a bit, but we'll have to deal with it just a while longer.

duk3luk3:
Hiding games that are more than 12 hours old is still a good idea. And I don't see why we'd have to revert that.

Wesmania:
I don't think so, certainly not without giving players a toggle option for that. That wouldn't be a problem anyway if the server didn't have the zombie game bug, I suggest we fix that instead.

GrotheFAF:
Currently the server issue is there, and it has effect in the client, a temp fix is better than no fix.
Come back, when you fixed that server issue. ('we fix' doesn't do anything.)
(and toggle option? really: like an option for: yes/no: want shit on my shoes ;-)

GrotheFAF:
a snapshot of the difference:
![fafclient-chat-user-with out -zombie](https://cloud.githubusercontent.com/assets/19195504/25509462/7c3b66f0-2bb9-11e7-921c-4451680107c0.jpg)

GrotheFAF:
Are 'we' happy with the current server solution? Is it working as expected?
So that I can close this?

Wesmania:
I believe that the server issue is partially fixed- we definitely no longer lose games forever. I've noticed that there's still another kind of issue - some games stay open as long as a player that played them stays online. Whether this is caused by a bug is left to investigate. It might be just players that don't close their FA, or we might be losing the 'FA closed' message from the client in one way or another.

In the first case I think the smart way to fix it would be to tell the client about individual players leaving the game - correct me if I'm wrong, but I think right now we determine whether the player is playing the game using the game state messages, so players that left a game early are marked as 'playing' until the server decides the game ended. It'd be a significant change since it involves changing server protocol, but it should be possible to add it while keeping backwards compatibility.

[...]

PR #705, opened 2017-04-22
fix for user in chat userlist shown in long gone games

GrotheFAF:
filter zombie games in client connection.py handle_game_info
workaround for #704 

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@
 * Clean up layout of login widget (#652, #653)
 * Use "foo" instead of user password in develop mode (#712, #653)
 * Remove unused label from GamesWidget (#717)
+* Fix for user in chat shown in long gone games (#705) 
 
 Contributors:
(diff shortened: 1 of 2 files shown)


Wesmania:
Can you move this code to src/client/connection.py, class LobbyInfo, method handle_game_info? This way we'll be able to kill all messages about old 'playing' games at the source.

We're also merging this one for the next release.

GrotheFAF:
done. 
(but it's not really 'at the source', just one function call earlier ;-)
(I used unused 'import config' for 'import time' hope you don't mind)

Wesmania:
Cool, do the usual rebase & changelog & we'll merge it in (I guess we leave #704 open until we fix it on the server).

GrotheFAF:
rebased and add changelog

PR #707, opened 2017-04-23
add host and live-game-delay to user status info in Chat Lobby

GrotheFAF:
The crossed sword will be red for the live delay time (5 minutes) and then turn grey.
The Tooltip gives title (for hosted games) and map info and live delay hint (during red sword).
The reversed single sword is shown for the game host.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


src/chat/chatter.py
```diff
@@ -184,11 +184,25 @@ def update(self):
...
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+                game = client.instance.games.games[int(url.path()[0:7])]  # get game_uid from url to get game
+                if time.time() - game.launched_at > 5 * 60:
+                    self.statusItem.setIcon(util.icon("chat/status/playing.png"))
+                    self.statusItem.setToolTip("Playing &nbsp;" + game.mod + "&nbsp; on &nbsp;" + game.mapdisplayname
+                                               + "<br/>" + url.toString())
+                else:
+                    self.statusItem.setIcon(util.icon("chat/status/playing_delay.png"))
+                    self.statusItem.setToolTip("Playing &nbsp;" + game.mod + "&nbsp; on &nbsp;" + game.mapdisplayname
+                                               + "&nbsp; - &nbsp;LIVE DELAY (5 Min)<br/>" + url.toString())
```
Wesmania, line 205: This repetition is pretty disgusting. Maybe we can move these icon names / tooltip generation to some tables and functions?

src/chat/chatter.py
```diff
@@ -184,11 +184,25 @@ def update(self):
...
+                    self.statusItem.setToolTip("Hosting Game Lobby: &nbsp;" + game.title + "&nbsp; on &nbsp;"
+                                               + game.mapdisplayname + "<br/>" + url.toString())
+                else:
+                    self.statusItem.setIcon(util.icon("chat/status/lobby.png"))
+                    self.statusItem.setToolTip("In Game Lobby: &nbsp;" + game.title + "&nbsp; on &nbsp;"
+                                               + game.mapdisplayname + "<br/>" + url.toString())
             elif url.scheme() == "faflive":
-                self.statusItem.setIcon(util.icon("chat/status/playing.png"))
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+                game = client.instance.games.games[int(url.path()[0:7])]  # get game_uid from url to get game
```
Wesmania: This is ugly and will break when game numbers roll over. We *need* to get that uid / game from elsewhere.

src/chat/chatter.py
```diff
@@ -184,11 +184,25 @@ def update(self):
         url = client.instance.urls.get(player.login)
         if url:
             if url.scheme() == "fafgame":
-                self.statusItem.setIcon(util.icon("chat/status/lobby.png"))
-                self.statusItem.setToolTip("In Game Lobby<br/>"+url.toString())
+                game = client.instance.games.games[int(url.queryItemValue("uid"))]
```
Wesmania, line 187: Same argument, it works, but it's really ugly. We need to get that uid from elsewhere.

src/games/gameitem.py
```diff
@@ -120,6 +121,9 @@ def url(self, player_id=None):
         
     @QtCore.pyqtSlot()
     def announceReplay(self):
+
+        client.instance.usersUpdated.emit(list(self.players))  # red to white crossed sword
```
Wesmania, line 125: Comment could be a bit more clear.

src/games/gameitem.py
```diff
@@ -285,7 +289,11 @@ def update(self, message, old_client=None):
         # Spawn announcers: IF we had a gamestate change, show replay and hosting announcements
         if oldstate != self.state:
             if self.state == "playing":  # The delay is there because we have a 5 minutes delay in the livereplay server
-                QtCore.QTimer.singleShot(5*60000, self.announceReplay)
+                self.launched_at = message['launched_at']  # needed in Chatter update for delay icon
+                if time.time() - self.launched_at > 5*60:  # most games after client start
+                    self.announceReplay()
+                else:
+                    QtCore.QTimer().singleShot(5*60010 - 1000*(time.time() - self.launched_at), self.announceReplay)
```
Wesmania, line 296: I think I've seen that code added lately in replays somewhere, maybe we can unify that somehow? Also why the 60010?

Wesmania (changes requested):
Code is ugly in places IMO, we need to refactor some of the existing mess if we are to add more features.

src/chat/chatter.py
GrotheFAF: Well, if you know another 'place', let me know. (and this is not a beauty contest)

src/chat/chatter.py
GrotheFAF: yes it will break, in a view years.
I can handle that right now, would be just a bit more code.

src/chat/chatter.py
GrotheFAF: 'disgusting' is the right word here?
You suggest more complex changes, I try to keep changes to a minimun, because of the code police...

src/games/gameitem.py
GrotheFAF: same argument as above, not to many changes at once.
the extra 0,1 second: during testing I found cases, where to singleshot seems to have come to early and the icon stayed red. with the extra time it's on the save side.

src/chat/chatter.py
Wesmania: We have a player just above, maybe we can use it for something? Worst case scenario we can pass the game or uid as an argument, but extracting the uid from the URL is unacceptable.

src/chat/chatter.py
Wesmania: Manually pulling arbitrary 8 chars from an url to extract a value somebody else put here is terrible code smell. One day we will maybe change something in the URL, or maybe move it somewhere, and this hack will cost us rippes hair.

src/chat/chatter.py
Wesmania: I had feature commits with significant refactors go in, and have plenty more pretty much approved. This one is really minor, get your act together.

src/games/gameitem.py
Wesmania: It sounds like it's a band-aid on a race condition. The bug can easily come back, since I'm sure timers don't give any guarantees on accuracy, and small UI freezes can probably skew them. I'd suggest calling this code from the replay tab, where there is 'show' logic already.

GrotheFAF:
I give up.

src/chat/chatter.py
```diff
@@ -184,11 +184,25 @@ def update(self):
...
+                    self.statusItem.setToolTip("Hosting Game Lobby: &nbsp;" + game.title + "&nbsp; on &nbsp;"
+                                               + game.mapdisplayname + "<br/>" + url.toString())
+                else:
+                    self.statusItem.setIcon(util.icon("chat/status/lobby.png"))
+                    self.statusItem.setToolTip("In Game Lobby: &nbsp;" + game.title + "&nbsp; on &nbsp;"
+                                               + game.mapdisplayname + "<br/>" + url.toString())
             elif url.scheme() == "faflive":
-                self.statusItem.setIcon(util.icon("chat/status/playing.png"))
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+                game = client.instance.games.games[int(url.path()[0:url.path().find("/")])]  # uid from url to get game
```
Wesmania, line 197: Is this supposed to be better?

PR #710, opened 2017-04-25
add host and live-game-delay to user status info in Chat Lobby

GrotheFAF:
add launched_at to GameItem to calculate announcement delay in Chatter
add QueryItem uid to url.scheme 'faflive'
use other single sword for host of hosted game
add title (only hosted), mod and map info to user status tooltip
remove url from user status tooltip
add playing_delay.png for live-game delay status icon
add status_unclear.png for user with url but no matching live game
remove unused status icon none.png

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


src/chat/chatter.py
```diff
@@ -183,12 +183,27 @@ def update(self):
...
         url = client.instance.urls.get(player.login)
         if url:
-            if url.scheme() == "fafgame":
-                self.statusItem.setIcon(util.icon("chat/status/lobby.png"))
-                self.statusItem.setToolTip("In Game Lobby<br/>"+url.toString())
-            elif url.scheme() == "faflive":
-                self.statusItem.setIcon(util.icon("chat/status/playing.png"))
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+            if int(url.queryItemValue("uid")) in client.instance.games.games:
+                game = client.instance.games.games[int(url.queryItemValue("uid"))]
```
Wesmania: I still don't like the way we get the uid here, but from what I've seen browsing the code lately, the very way urls find their way into here is quite a mess in itself. It can probably work for now, especially given that we pull things from the url right above it anyway. One day we'll probably refactor the url stuff.

src/chat/chatter.py
```diff
@@ -183,12 +183,27 @@ def update(self):
...
+                elif url.scheme() == "faflive":
+                    if time.time() - game.launched_at > 5 * 60:
+                        self.statusItem.setIcon(util.icon("chat/status/playing.png"))
+                        self.statusItem.setToolTip("Playing" + game_str)
+                    else:
+                        self.statusItem.setIcon(util.icon("chat/status/playing_delay.png"))
+                        self.statusItem.setToolTip("Playing" + game_str + " -  LIVE DELAY (5 Min)")
+            else:  # that shouldn't happen - but it does in rare cases
+                self.statusItem.setIcon(util.icon("chat/status/status_unclear.png"))
+                self.statusItem.setToolTip("(has url - but no matching running game found)<br/>" + url.toString())
```
Wesmania, line 206: This became a lot of cases IMO (6 up from 3), we could try decoupling figuring out which state to choose from setting stuff up. Maybe something along the lines of:
```
class ChatterGameState(Enum):
    IDLE = ...
    HOST = ...
    LOBBY = ...
```
With some function like:
```
def setChatterGameState(self, state, url, game = None)
```
that does the setting depending on the state.
It's mostly nitpicking that could probably get fixed after other refactors, but would be nice to have.

src/games/gameitem.py
```diff
@@ -285,9 +290,13 @@ def update(self, message, old_client=None):
         # Spawn announcers: IF we had a gamestate change, show replay and hosting announcements
         if oldstate != self.state:
             if self.state == "playing":  # The delay is there because we have a 5 minutes delay in the livereplay server
-                QtCore.QTimer.singleShot(5*60000, self.announceReplay)
+                self.launched_at = message['launched_at']
+                if time.time() - self.launched_at > 5*60:  # most already running games at client start
+                    self.announceReplay()
+                else:
+                    QtCore.QTimer().singleShot(5*60100 - 1000*(time.time() - self.launched_at), self.announceReplay)
```
Wesmania: This works for now, but we definitely need a comment explaining where that 60100 came from. Honestly all our woes from the timer come from the fact that we really need to have a single place away from any widgets that announces replay appearance.

I did some work in #709 that centralizes processing games from the server - it's some broad and destructive refactoring that can probably be merged no sooner than 0.14, but if it does, we'll be able to make a single class that reports live replays after a delay and just hook up to it.

src/games/gameitem.py
```diff
@@ -285,9 +290,13 @@ def update(self, message, old_client=None):
...
             elif self.state == "open":  # The 3.5s delay is there because the host needs time to choose a map
-                QtCore.QTimer.singleShot(35000, self.announceHosting)
+                QtCore.QTimer().singleShot(35000, self.announceHosting)
```
Wesmania: Qt docs tell me that singleShot is a static method, and that instance singleShot is a bool attribute, so I think the original was correct.

GrotheFAF: well, both versions seem to work. (I can't tell the difference)
Without the () PyCharm complain about the time number ('Expected type 'QTimer', got 'int' instead'), with the () no complain. So from PyCharm perspective with the () is better.
Then again, PyCharm isn't the wisest in Qt things.

src/games/gameitem.py
GrotheFAF: Yes, it's strange to have the central gamelist in QListWIdgetItem in a QListWidge (with most Items hidden).
Better to have a Game and Games class, like Player and Players. (like you've done in #709, I would prefer it in client folder, but **if we move Player/Players to model[s] folder too** maybe .. cause 'model' doesn't sound like games, so I wouldn't look there for it .. naming logic ... hard stuff ;-) 
I just recently noted, that the server message is handed over from Widget to Widget (game replay coop). 
In the end The GamesWidget should only hold the hosted games.

src/games/gameitem.py
Wesmania: I guess it complains because it thinks we're using the method directly and expects 'self'. Qt itself says it's a static method - I say we keep using it as a static method, PyCharm shenanigans aside.

src/games/gameitem.py
Wesmania: Well, there's also the coop widget which is also interested in hosted games, so I feel having a dedicated class to keep track of games is a nice idea. I called the folder 'model', because I envision it containing the 'model' classes like Player, Game, Mod etc., and other folders are named more after a specific tab than a specific functionality.
And yeah, the games widget should remove the games that go to 'playing' state, that's still to be done.

GrotheFAF: so we agree (see **bold** above).

src/games/gameitem.py
GrotheFAF: no, not the self thing, PyCharm is just unhappy without the ().

src/games/gameitem.py
Wesmania: Yeah, I'd like to see it happen later. We can move it (refactoring if necessary) in a future PR.

src/games/gameitem.py
Wesmania: I meant that `QTimer.singleShot(timer, ...)` is equivalent to `timer.singleShot(...)` if `singleShot` is a normal method. PyCharm seems to get confused and thinks that `singleShot` is a normal, not static method, so it expects its invocation will have QTimer as the first argument. Maybe PyQt has some automatic binding generation that confuses PyCharm, but it's certainly in the wrong here.

Wesmania: It seems to be a known issue in PyCharm / PyQt, too:
https://youtrack.jetbrains.com/issue/PY-14208#tab=Linked%20Issues

GrotheFAF: removed the ()

src/chat/chatter.py
GrotheFAF: plus cases for PLAYING5, PLAYING & STRANGE (has url but no matching game).
and you still need a 'case study' to evaluate and set ChatterGameState, that would be in setChatterGameState I guess. 
and would that Chattergamestate then be stored?
the chatter has a [game]status
would the Icon and tooltip (a bunch of Formatter I asume) be set in that class?

one could put icon and tooltip in strings and do the 2 self.StatusItem.set... once at the end.

PR #717, opened 2017-04-28
remove unused label [subset]RankHint from gameswidget

GrotheFAF:
There are 2 RankHint labels with slightly different text, only one 'subsetRankHint' was used. I removed the the older unused "RankHint" and renamed 'subsetRankHint' to 'RankHint'.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@
 * Fix crash on opening files in explorer on non-Windows (#654, #655)
 * Clean up layout of login widget (#652, #653)
 * Use "foo" instead of user password in develop mode (#712, #653)
+* Remove unused label from GamesWidget (#717)
 
 Contributors:
   - Wesmania
--- a/res/client/client.css
+++ b/res/client/client.css
@@ -434,7 +434,7 @@ QLabel#labelSortGames,#labelJoinGame,#labelHostGame,#labelAutomatch,#labelMyRepl
     color:white;
 }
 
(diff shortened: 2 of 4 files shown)


Wesmania:
I like it. It sucks that designer is so git-unfriendly - it seems it randomly decided to remove a single space everywhere, replacing 600+ lines. I don't know if it'd be more of a pain to roll with a full replacement this once and clobber commit history, or to manually fix commits (doesn't seem especially bad in this case, we can script adding a single space everywhere).

Wesmania:
I tested it a tiny bit - it seems the actual problem is that qt designer decided it likes 1-space indentations rather than 2-space ones,and that causes changes everywhere. I have absolutely no idea if it's configurable, or if our .ui files are even consistent about it. If it turns out we need to change that indentation, it might be best to split this commit into one that only changes indentation and one that makes meaningful changes.

Wesmania:
I did some more checking - all the ui files except for this file use 1-space indents. We definitely want to bring it in line with the rest - maybe split the indent change to its own commit, so we can more easily see what meaningful changes are here. The indent thing is also worth putting on the wiki.

Wesmania:
Rebase on develop (+ changelog), maybe add a separate commit just for indent change, and we'll merge it in.

GrotheFAF:
rebased and add changelog

PR #719, opened 2017-04-30
718 refactor vault downloads

Wesmania:
This code refactors downloading assets from the vault. It adds the following:

- FileDownload class for convenient download of files from a url,
- VaultDownloadDialog for presenting a modal dialog while downloading a file
- downloadVaultAsset function that uses both of the above to download a file, do logging and present other modal windows to the user,
- refactored downloadMap and downloadMod that use the above.

These changes should make adding support for downloading tutorials easy.

I'd like comments on the FileDownload class - there's some complex interactions with QNetworkReply signals that feel non-obvious and seem hard to unit test.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@
 * Add Friend/Foe feature for IRC users (#609, #657)
 * Replace game quality in the play tab (NOT in the lobby) by average rating (#687, #688)
 * Split replay widget to separate classes (#698, #699)
+* Cleanup and unify map & mod download behaviour (#718, #719)
 
 Contributors:
  - Wesmania
--- a/src/downloadManager/__init__.py
+++ b/src/downloadManager/__init__.py
@@ -119,52 +119,41 @@ class downloadManager(QtCore.QObject):
         self.client = parent
         self.nam = QNetworkAccessManager(self)
 
-        self.nam.finished.connect(self.finishedDownload)
-
         self.modRequests = {}
         self.mapRequests = {}
         self.mapRequestsItem = []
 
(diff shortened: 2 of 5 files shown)


Brutus5000:
I don't know whether it's a specific Qt4 thing, but the downloader seems to do quite unnecessary low level stuff. I.e. do we care about the number of redirects? Qt should notice an infinite loop.
For a minimal Qt5 downloader see [here](https://github.com/Brutus5000/BiReUS-Launcher/blob/master/launcher/download_service.py).

I'd suggest using signals to broadcast any changes of the downloader so you can listen to the events from anywhere (and any thread). Decoupling a service using signal/slots is one of the core ideas behind Qt.

duk3luk3:
> I don't know whether it's a specific Qt4 thing, but the downloader seems to do quite unnecessary low level stuff. I.e. do we care about the number of redirects? Qt should notice an infinite loop.

Qt4 NetworkAccessManager doesn't handle redirects.

Wesmania:
Thanks for the link to the downloader brutus, I see a couple of thing that will simplify stuff a lot (didn't notice that downloadProgress signal, for example).

Wesmania:
We can simplify FileDownload code by a lot if we don't bother with redirects just yet. The original code used requests, and it follows redirects, but it's only used to access files in the vault, and I haven't seen any redirects there. I think we can stomach no redirect support until we get to qt5.

duk3luk3:
As soon as we do something sensible on the server like enforce https, there will be redirects.

PR #730, opened 2017-05-08
feature - singleline the layout of Vaults\Mods

GrotheFAF:
This will put all textboxes and buttons in a single line and cleanup the Layout.

QtDesigner:
before:
fafclient-modvault-ui before
after:
fafclient-modvault-ui after

FAFclient:
before:
fafclient-mods-elements before
after:
fafclient-mods-elements after

### DIFF
--- a/res/modvault/modvault.ui
+++ b/res/modvault/modvault.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>625</width>
+    <width>729</width>
     <height>376</height>
    </rect>
   </property>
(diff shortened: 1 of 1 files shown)


duk3luk3:
Can you please also make the enter key work on the search field?

GrotheFAF:
would 'enter key word[s]'  be, ok? ..test.. test...
Not ok, just one, more than one gives nothing.
So to make that clear 'enter one key word' ?
or maybe 'enter (just one) key word'

duk3luk3:
I mean that hitting the "enter" ("return") key on the keyboard while focused on the search field should do the same thing as clicking the search button.

GrotheFAF:
It doesn't?
May I refer you to:
https://github.com/FAForever/client/blob/develop/src/modvault/__init__.py#L87 + L88

CookieNoob:
you can also remove/hide the upload mod button. Its broken since the server change.

PR #742, opened 2017-06-04
Log errors from faf-uid executable

Wesmania:
faf-uid uses stderr to communicate errors. Use Popen to read stdout and
stderr separately, then use stderr to print faf-uid errors instead of
stdout.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,9 +3,11 @@
 
 * Single-Line Mod-Vault toolbar (#730)
 * Pin UID version in appveyor
+* Log errors from faf-uid executable (#741, #742)
 
 Contributors:
   - Grothe
+  - Wesmania
 
 
 0.12.5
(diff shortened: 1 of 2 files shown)


duk3luk3:
Seems to work as far as I've been able to test it (with no faf-uid executable, with just an empty unexecutable file as faf-uid executable, and with a secret faf-uid executable that has a broken rsa key).

Need to verify it works on Windows too, which can be weird about FD's - I will do that asap.

But why not just pass `stderr=subprocess.STDOUT` to `check_output`? (https://docs.python.org/2/library/subprocess.html#subprocess.STDOUT)

Wesmania:
I didn't want to do that since it melds stderr and stdout together. I don't want to rely on faf-uid writing only on either stdout or stderr, but we can rely on one being used for errors, and another for output.

duk3luk3:
Oh, of course, I totally forgot about that! Please rebase and I'll merge.

PR #748, opened 2017-06-09
Cleanup resize behaviour for client / channel

Wesmania:
I think there was once an attempt of optimization, with resize signals
from the client. Thet definitely don't work since the client eats a lot
of CPU while resized anyway, signals are only used in one place and
cause some timers to potentially get called a hundred times a second,
Let's clean this up and worry about resize performance some other day.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 * Pin UID version in appveyor
 * Log errors from faf-uid executable (#741, #742)
 * Add menu option: Show paths for maps, mods & replays (#735)
+* Slightly clean up client resize logic (#747, #748)
 
 Contributors:
   - Grothe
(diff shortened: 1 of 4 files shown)


GrotheFAF:
This changes startup chat tabs from:
Server / #aeolus / #..clan
to
Server / #...clan / #aeolus

this can be fixed in _chatwidget with:
L280 +        self.channels[channel] = newch
L287 +     else:

Wesmania:
You're right - dammit, I missed that. I can see why it happens, I'll fix it in a follow-up PR.

GrotheFAF:
and while you're at it...
in on_join 'channel' is the channel-name[string], in addChannel 'channel' is the channel-object.
That is confusing naming. (wonder why the code police let that fly...)
(and 'newch' .. what happen to 'new_channel', was that taken?-)

Wesmania:
C naming habits, sorry. I'd rather not do renaming in commits that change logic, we can make some trivial rename / reorder PRs later.

PR #751, opened 2017-06-13
Fix wrong channel order caused by 5a8ac3c4a1ec

Wesmania:
#748 accidentally changed chat tab order. This PR fixes it.

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -128,10 +128,12 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
                     self.channels[channel].chatters[player].avatarItem.setIcon(QtGui.QIcon(util.respix(reply.url().toString())))
                     self.channels[channel].chatters[player].avatarItem.setToolTip(self.channels[channel].chatters[player].avatarTip)
 
-    def addChannel(self, name, channel):
+    def addChannel(self, name, channel, index = None):
         self.channels[name] = channel
-        self.addTab(self.channels[name], name)
-        channel.setTextWidth()
+        if index is None:
+            self.addTab(self.channels[name], name)
+        else:
+            self.insertTab(index, self.channels[name], name)
 
     def closeChannel(self, index):
         """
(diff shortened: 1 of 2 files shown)


GrotheFAF:
looks good and works as expected.   ... and found another problem - see below.
(and ignore my reverence above, just took your changes in for testing ;-)

(but you might change the PR-title: commit-Nr to fix-PR-Nr)

GrotheFAF:
found another problem: 
Calling !smurfcheck [name] in #aeolus ->  adds TheSetoner channel  - with single-char width per line.
This is also true for joining a #clan channel
it's ok before this commit.

Wesmania:
Seems hidden widgets don't have their size set, I pushed a fix for that. Setting text width when we're shown is probably the smartest, and makes sense since I don't even know if hidden widgets process resize events anyway.

GrotheFAF:
that seems to do it... 
but...
if I click in Chat to change a channel, I get:
resizeEvent .. setTextWidth() - showEvent ..setTextWidth() - resizeEvent ..setTextWidth()
that is a lot of (setTextWidth())
- For testing I added a 'if self.isVisible():' before each setTextWidth(), I noticed on the above channel change, the first resizeEvent isVisible() is false, the subsequent showEvent and 2nd resizeEvent isVisible() is true.
But entering a new channel after !smurfcheck or /join, only has resizeEvent - showEvent (no 2nd resizeEvent), with the setTextWidth() in 1rst resizeEvent having no effect (isVisible being false)
so I guess that's how it is...
(resizeEvent  is also called on ladder-lure show and hide)
(and you lost the 'to no effect'-setTextWidth in addChannel, good ;-)
I asume now it's clearer what the timer was for, isn't it?-)
tricky... but this still looks better.

Wesmania:
I don't think we should worry about resize events happening too often, I trust Qt to behave sanely in that regard, and it indeed seems to. If we ever need to throttle resize events, we'll do it properly.

PR #760, opened 2017-06-19
fix client frameless, missing import and reference

GrotheFAF:
The client wasn't framless - fixed that with setWindowFlags(QtCore.Qt.FramelessWindowHint)
In scr/__init__ 2 references were missing.


Wesmania:
If you make this 3 separate commits, you'll make me a very happy rebaser <3

Wesmania:
Making this 3 separate changes would make it much easier to rebase and meld together related changes later.

Wesmania:
I just cherrypicked and split it into the branch. We could work like this in the future instead of making PRs.

Issue #540, opened 2016-12-23
"Host has left the game" message when trying to join a game


JaggedAppliance:
**I experienced...** As the title states, an a couple of occasions I've gotten this message when trying to join a lobby. It doesn't join, just gives me that message.

https://github.com/FAForever/fa/files/517790/forever.zip
https://github.com/FAForever/fa/files/670022/forever.zip

Edit: I'm using the 0.12.0 preview 2

duk3luk3:
This happens with 0.11.60 too, I assume? Since you opened the original issue on Oct. 8...

JaggedAppliance:
Yes it does.

duk3luk3:
Does it result in not being able to join that lobby at all? I.e. if it happens once, will it always happen for that lobby?

JaggedAppliance:
Yes, in my experience if it happens it continues to happen every time I try to join that lobby.

duk3luk3:
More info from aeolus:

```
16:08:48 < goodgue> it happens whenever me and my friend try to enter a lobby we created
16:09:17 < goodgue> but otherwise i can go into anyone elses lobby
16:09:45 < OppressiveDuke> does it happen no matter which of you created the lobby?
16:09:53 < goodgue> yeah
16:10:04 < OppressiveDuke> ok I'll make a note of that
16:10:18 < goodgue> however, last night we somehow managed to get into a lobby - but when we tried it again afterwards it gave the same error
```

JaggedAppliance:
[forever (2).zip](https://github.com/FAForever/client/files/732898/forever.2.zip)
happened again, more logs

duk3luk3:
Which game mods does this happen with?

It's probably https://github.com/FAForever/server/issues/197, which is a time-out. It seems to mostly happen with coop. Does it happen with other game mods for you?

JaggedAppliance:
For me it happened with the normal faf mod.

duk3luk3:
OK, thanks!

GrotheFAF:
Is this independed of joining via user in Chat or joining via game in Play/Coop?

speed2CZ:
This is fixed and can be closed

PR #796, opened 2017-07-18
fix join hosted game from chat user

GrotheFAF:
sending uid as string didn't work anymore (unclear since when), packed as int().
might be related to #540


Wesmania:
This is a bug that's also in stable, so please add a changelog entry.

GrotheFAF:
added changelog

PR #797, opened 2017-07-18
fix live game ended online replay doubleclick

### LINKED ISSUE #783: finished live replays cant be loaded
**I experienced...** this error in the stable release 0.13.2+2070:
grafik

**how to reproduce** search for replays in the vault and find a live game (in "playing" state), then wait until the game ends and try to launch the replay without refreshing the list

**I expected...** loading the normal replay file

**Where the bug is / What should be done** handle the finished live games correctly

GrotheFAF:
if a search gets old, shown 'playing' games might already have ended, 
which leads to a bad click experience, we don't want that.

check if live-game is still running, else give option to start replay, fix #783
check for (a) player still in clicked game (need to add uid to url)
add error message if replay ended irregular

### DIFF
--- a/src/model/game.py
+++ b/src/model/game.py
@@ -180,11 +180,11 @@ class Game(QObject):
         query = QUrlQuery()
         query.addQueryItem("map", self.mapname)
         query.addQueryItem("mod", self.featured_mod)
+        query.addQueryItem("uid", str(self.uid))
 
         if self.state == GameState.OPEN:
             url.setScheme("fafgame")
             url.setPath("/" + str(player_id))
-            query.addQueryItem("uid", str(self.uid))
         else:
             url.setScheme("faflive")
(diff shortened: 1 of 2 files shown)


Wesmania:
The logic is really convoluted, I have absolutely no idea if it's doing the right thing. We'd also really like to move away from the old coding style of 'let's pull the data we need from wherever and increase the size of our functions forever'. Since it's not a common bug, and using url for that info is a hack anyway, and since I'm working on the new object model, I suggest waiting on this one and redoing it the right way in the future.

GrotheFAF:
I don't 'really' understand you refactors 'absolutely' neither .. so maybe make an effort.
If you 'really' know a better way to get the data, please show me.
The url was there before, the new things are the warnings and question msgboxes.
leaving bugs unhandled because the code would look bad, sounds 'really' stupid.
And you 'suggest'? I have an option? (words have meaning)

PR #801, opened 2017-07-21
[WIP] fix map preview

muellni:
while trying to reproduce http://forums.faforever.com/viewtopic.php?f=2&t=4507&start=320#p152255 I deleted my cache folder content and noticed that no previews are generated for default scmp-maps.

On my retail installation the scmp-maps are stored uppercase on disk and this little patch made the preview work. 
But I don't know if this breaks  some other thumbnail stuff later.

### DIFF
--- a/src/fa/maps.py
+++ b/src/fa/maps.py
@@ -279,13 +279,13 @@ def __exportPreviewFromMap(mapname, positions=None):
         mapdir = mapname
     elif os.path.isdir(os.path.join(getUserMapsFolder(), mapname)):
         mapdir = os.path.join(getUserMapsFolder(), mapname)
-    elif os.path.isdir(os.path.join(getBaseMapsFolder(), mapname)):
-        mapdir = os.path.join(getBaseMapsFolder(), mapname)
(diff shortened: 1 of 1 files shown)


Wesmania:
I could merge this, did you check if thumbnail stuff indeed doesn't get broken?

muellni:
The only thing I checked was that I got a thumbnail for Open Palms+Setons with this patch and for custom maps. But I only checked on Linux. I didn't check
* anything on Windows
* the steam version of FA on Linux

muellni:
I just did some testing on windows:
Py3Qt5 client 0.15.1 beta starts, but hangs with max CPU and eats memory. It does generate some previews in the cache directory, but only for standard maps (scmp/x1mp), all other maps seem to have 0kb-files created for them.
I'm going to check this again with d3d acceleration disabled

muellni:
with current develop branch, this issues seems to be fixed, I'll rebase and check again

duk3luk3:
Any news?

muellni:
It works now on `develop`, so I think this fix isn't needed anymore.

PR #809, opened 2017-07-23
beautify and correct everything

GrotheFAF:
clear and correct blanks and lines
removing unused & add missing imports
rename some vars and constants, Camel some classes


GrotheFAF:
4% that is something ;-)

GrotheFAF:
split it

PR #816, opened 2017-07-31
[RAT] refactor chatter pressed

GrotheFAF:
lose the setDisabled() - insteat add only needed menu topics with menu_add()
factor out player/irc into f_action_connect()

(sadly there is no permanence in me.addF../remF..)

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -265,127 +265,79 @@ class Chatter(QtWidgets.QTableWidgetItem):
     def pressed(self, item):
         menu = QtWidgets.QMenu(self.parent)
 
-        # Actions for stats
-        actionSelectAvatar = QtWidgets.QAction("Select Avatar", menu)
-
-        # Action for aliases link
-        actionViewAliases = QtWidgets.QAction("View Aliases", menu)
-
-        # Actions for Games and Replays
-        actionReplay = QtWidgets.QAction("View Live Replay", menu)
-        actionVaultReplay = QtWidgets.QAction("View Replays in Vault", menu)
-        actionJoin = QtWidgets.QAction("Join in Game", menu)
-
-        # Default is all disabled, we figure out what we can do after this
-        actionReplay.setDisabled(True)
-        actionJoin.setDisabled(True)
-
-        # Don't allow self to be invited to a game, or join one
-        if self.lobby.client.login != self.name:
-            if self.name in client.instance.urls:
-                url = client.instance.urls[self.name]
-                if url.scheme() == "fafgame":
-                    actionJoin.setEnabled(True)
-                elif url.scheme() == "faflive":
-                    actionReplay.setEnabled(True)
-
-        # Triggers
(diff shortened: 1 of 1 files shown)


src/chat/chatter.py
```diff
@@ -265,127 +265,79 @@ def doubleClicked(self, item):
...
-                elif url.scheme() == "faflive":
-                    actionReplay.setEnabled(True)
-
-        # Triggers
-        actionViewAliases.triggered.connect(self.viewAliases)
-        actionSelectAvatar.triggered.connect(self.selectAvatar)
-        actionReplay.triggered.connect(self.viewReplay)
-        actionVaultReplay.triggered.connect(self.viewVaultReplay)
-        actionJoin.triggered.connect(self.joinInGame)
+        def menu_add(action_str, action_connect, separator=None):
```
Wesmania: If separator is a bool, we could make if default to False.

src/chat/chatter.py
```diff
@@ -265,127 +265,79 @@ def doubleClicked(self, item):
...
-                actionCloseLobby = QtWidgets.QAction("Close FAF Client", menu)
-                menu.addAction(actionCloseLobby)
-                actionCloseLobby.triggered.connect(lambda: self.lobby.client.closeLobby(self.name))
+        # Replays in vault
+        if self.id != -1:  # not for irc user
+            menu_add("View Replays in Vault", self.viewVaultReplay, True)
 
-            menu.addSeparator()
+        # Friends and Foes Lists
+        def f_action_connect(f, irc_f):  # Irc or not Irc, that's the Question
```
Wesmania: I don't know if f_action_connect is descriptive enough, how about player_or_irc_action?

Wesmania (commented):
Looks really good, just a nitpick or two.

src/chat/chatter.py
GrotheFAF: it is the action_connector, got the f_ (Friends_/Foes_) to distinguish from the one in menu_add.

Wesmania: I know what it does, I just feel that one can't tell what it does straight from its name. If we change it, we probably won't even need the comments explaining what it does. I think we even have some code guideline about such naming buried in client readme.

Wesmania:
What do you mean by permanence? If it's some feature, we could go ahead and add an issue for it.

GrotheFAF:
The user add/remFriend/Foe changes don't survive a Client restart.
(the clientwindow versions do, and the user add/remIrcFriend/Foe do)

src/chat/chatter.py
GrotheFAF: I like the comment.

GrotheFAF: renamed it to player_or_irc_action_connect

src/chat/chatter.py
GrotheFAF: I had False at first, but None makes more sense.

GrotheFAF: renamed it to False

PR #818, opened 2017-08-01
Beautify code

GrotheFAF:
fixing blanks and line breaks and line conventions


duk3luk3:
please squash that into a single commit - and it's "indenting", not "intending".

and add changelog if you want.

GrotheFAF:
done

PR #822, opened 2017-08-08
Improve client updater UX

duk3luk3:
* Never show same version client already has as update
* Don't call old versions "updates"
* Don't show old versions by default

### DIFF
--- a/src/client/updater.py
+++ b/src/client/updater.py
@@ -22,23 +22,28 @@ class UpdateSettings:
 
     def should_notify(self, releases, force=True):
         self._logger.debug(releases)
-        have_server = 'server' in releases
-        have_stable = 'stable' in releases
-        have_pre = 'pre' in releases
-        have_beta = 'beta' in releases
-
         if force:
-            return have_server or have_stable or have_pre or have_beta
-        else:
-            current_version = Version(config.VERSION)
-            # null out build because we don't care about it
-            current_version.build = ()
+            return True
 
-            notify_stable = have_stable and self.updater_branch == UpdateBranch.Stable.name and Version(releases['stable']['new_version']) > current_version
-            notify_pre = have_pre and self.updater_branch == UpdateBranch.Prerelease.name and Version(releases['pre']['new_version']) > current_version
-            notify_beta = have_beta and self.updater_branch == UpdateBranch.Unstable.name and Version(releases['beta']['new_version']) > current_version
+        server_releases = [release for release in releases if release['branch']=='server']
+        stable_releases = [release for release in releases if release['branch']=='stable']
+        pre_releases = [release for release in releases if release['branch']=='pre']
+        beta_releases = [release for release in releases if release['branch']=='beta']
 
-            return have_server or notify_stable or notify_pre or notify_beta
+        have_server = len(server_releases) > 0
+        have_stable = len(stable_releases) > 0
+        have_pre = len(pre_releases) > 0
+        have_beta = len(beta_releases) > 0
+
+        current_version = Version(config.VERSION)
+        # null out build because we don't care about it
+        current_version.build = ()
+
+        notify_stable = have_stable and self.updater_branch == UpdateBranch.Stable.name and Version(stable_releases[0]['new_version']) > current_version
+        notify_pre = have_pre and self.updater_branch == UpdateBranch.Prerelease.name and Version(pre_releases[0]['new_version']) > current_version
+        notify_beta = have_beta and self.updater_branch == UpdateBranch.Unstable.name and Version(beta_releases[0]['new_version']) > current_version
+
+        return have_server or notify_stable or notify_pre or notify_beta
 
 FormClass, BaseClass = util.THEME.loadUiType("client/update.ui")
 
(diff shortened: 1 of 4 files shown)


src/client/updater.py
```diff
@@ -22,23 +22,28 @@ class UpdateSettings:
...
+
+        server_releases = [release for release in releases if release['branch']=='server']
+        stable_releases = [release for release in releases if release['branch']=='stable']
+        pre_releases = [release for release in releases if release['branch']=='pre']
+        beta_releases = [release for release in releases if release['branch']=='beta']
 
-            notify_stable = have_stable and self.updater_branch == UpdateBranch.Stable.name and Version(releases['stable']['new_version']) > current_version
-            notify_pre = have_pre and self.updater_branch == UpdateBranch.Prerelease.name and Version(releases['pre']['new_version']) > current_version
-            notify_beta = have_beta and self.updater_branch == UpdateBranch.Unstable.name and Version(releases['beta']['new_version']) > current_version
+        have_server = len(server_releases) > 0
```
Wesmania, line 33: We can as well rely on the fact that empty lists evaluate to false and do 'if server_releases'.

src/client/updater.py
```diff
@@ -22,23 +22,28 @@ class UpdateSettings:
...
-            return have_server or notify_stable or notify_pre or notify_beta
+        current_version = Version(config.VERSION)
+        # null out build because we don't care about it
+        current_version.build = ()
+
+        notify_stable = have_stable and self.updater_branch == UpdateBranch.Stable.name and Version(stable_releases[0]['new_version']) > current_version
+        notify_pre = have_pre and self.updater_branch == UpdateBranch.Prerelease.name and Version(pre_releases[0]['new_version']) > current_version
+        notify_beta = have_beta and self.updater_branch == UpdateBranch.Unstable.name and Version(beta_releases[0]['new_version']) > current_version
+
+        return have_server or notify_stable or notify_pre or notify_beta
```
Wesmania, line 46: I think we can rearrange that to use an if / if else / else with self.updater_branch (maybe add a 'version we're interested in' variable to make things easier). Also, are we supposed to use that 'downgrade' config variable here?

src/client/updater.py
```diff
@@ -71,30 +76,39 @@ def reset_controls(self):
...
         else:
-            self.lblUpdatesFound.setText('Client updates were found.')
+            self.lblUpdatesFound.setText('Client releases were found.')
 
         if len(self.releases) > 0:
             currIdx = 0
-            preferIdx = 0
+            preferIdx = -1
 
+            labels = dict([('server', 'Server Version'), ('stable', 'Stable Version'), ('pre', 'Stable Prerelease'), ('beta', 'Unstable')])
```
Wesmania: Not the prettiest notation IMO

src/client/updater.py
```diff
@@ -71,30 +76,39 @@ def reset_controls(self):
...
+        if any([release['branch']=='server' for release in self.releases]):
             self.lblUpdatesFound.setText('Your client version is outdated - you must update to play.')
         else:
-            self.lblUpdatesFound.setText('Client updates were found.')
+            self.lblUpdatesFound.setText('Client releases were found.')
 
         if len(self.releases) > 0:
             currIdx = 0
-            preferIdx = 0
+            preferIdx = -1
```
Wesmania: None would be nice too (just a personal preference / nitpick).

src/client/updater.py
```diff
@@ -71,30 +76,39 @@ def reset_controls(self):
...
-                        preferIdx = currIdx
+                is_update = ' [New!]' if Version(release['new_version']) > current_version else ''
 
-                    currIdx = currIdx + 1
+                self.cbReleases.insertItem(99, '{} {}{}'.format(label, release['new_version'], is_update), release)
+
+                if prefer and preferIdx == -1:
+                    preferIdx = currIdx
+
+                currIdx = currIdx + 1
```
Wesmania: We can just iterate over enumerate(self.releases) instead.

src/client/updater.py
```diff
@@ -141,13 +155,9 @@ def showSettings(self):
 @with_logger
 class UpdateChecker(QObject):
     gh_releases_url = Settings.persisted_property('updater/gh_release_url', type=str, default_value='https://api.github.com/repos/FAForever/client/releases?per_page=20')
+    updater_downgrade = Settings.persisted_property('updater/downgrade', type=bool, default_value=False)
```
Wesmania, line 156: Why is it in 2 places at once? Is persisted_property capable of handling such a case (specifically does it update the other value when one of them changes)?

src/client/updater.py
```diff
@@ -172,11 +182,34 @@ def server_session(self):
...
                     download_url = asset['browser_download_url']
                     tag = release_dict['tag_name']
+                    v = Version(tag)
+                    # We never want to return the current version itself,
+                    # but if `updater_downgrade` is set, we do return
+                    # older releases.
+                    # strange comparison logic is because of semantic_version
+                    # so that build info is ignored
+                    if self.updater_downgrade:
+                        if not v < cv or cv < v
```
Wesmania: Longer variable names would help readability here IMO, something like 'current_version' and 'remote_version'. And I think operator precedence screwed us over here.

Wesmania (changes requested):
Some nitpicks, one logic issue.

src/client/updater.py
duk3luk3: I prefer not to rely on falsyness.

src/client/updater.py
duk3luk3: * I don't see how to make the logic more concise without making it harder to understand. I don't think turning it into a cascade of if statements would help at all.
* No, this logic is meant for deciding whether we want to show the update dialog when the client starts.

src/client/updater.py
duk3luk3: persisted_property is just a wrapper around Settings.get and Settings.set so yes, this is fine.

src/client/updater.py
duk3luk3: what screwed us here is that I left off the ':'. The precedence [is fine](https://docs.python.org/3/reference/expressions.html#operator-precedence).

src/client/updater.py
Wesmania: It's fairly pythonic IMO, but your call.

src/client/updater.py
Wesmania: We can change notify_* bools to check whether a given branch has a newer version (remove  the self.updater_branch check), then use self.updater_branch separately.

src/client/updater.py
```diff
@@ -172,11 +180,34 @@ def server_session(self):
...
+                    # We never want to return the current version itself,
+                    # but if `updater_downgrade` is set, we do return
+                    # older releases.
+                    # strange comparison logic is because of semantic_version
+                    # so that build info is ignored
+                    if self.updater_downgrade:
+                        if not (release_version < client_version or client_version < release_version):
+                            return None
+                    else:
+                        if not (release_version > client_release_versionersion):
```
Wesmania: Typo?

PR #823, opened 2017-08-08
Fix instructions in README in order for FAF to find faf-uid on Linux

iczero:
FAF expects faf-uid to be in PATH

Initial PR: N/A, this is a PR to the readme

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -77,7 +77,7 @@ Note that the `faf-uid` smurf protection executable needs to run `xrandr`, `lspc
 
 Run the client:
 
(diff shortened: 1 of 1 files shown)


duk3luk3:
Thanks for your PR! I don't know if we'll merge your PR directly, @Wesmania might have something to say about paths.

Wesmania:
Sounds OK, giving this one a +2.

duk3luk3:
OK - @iczero please add a changelog entry if you want!

iczero:
Does this really need a changelog entry? It's just modifying the readme to be a bit clearer.

PR #826, opened 2017-08-10
Clean up whitespace with autopep8

Wesmania:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


GrotheFAF:
I guess this is obsolete after the refactor

Wesmania:
It is, but it's just a matter of running autopep8 again and pushing the changes. We can do this any time with no effort, and I don't want to do this just yet to avoid impeding the ICE branch.

ChrisKitching:
The problem with stuff like this is:
- It clobbers quite a bit of blame
- It will cause merge conflicts with almost all of the existing feature branches/PRs (of which there seem to be many)
- It doesn't really _achieve_ anything.

My suggestion would be to apply trivial formatting changes like this as you go along. When you touch a function, feel free to make its spaces "right" (having first agreed with everyone what arbitrary style rules we're using, so you don't end up in an edit war). This approach also has the advantage of avoiding the creation of a meaningless blame layer...

Sheeo:
I concur, we've historically denied PR's which do this for the above reasons.

PR #830, opened 2017-08-14
clarify some variables

GrotheFAF:
lose most import ... as ...
rename some variables clarify widgets
CamelCase classes


Wesmania:
If you don't mind, I'll pull this PR last (of the ones you made lately) - my new-object-model branch does some changes to chat widgets and I'd like to rebase it a bit later.

Issue #813, opened 2017-07-26
friend aren't all at the top of name list in aeolus


shalkya:
![image](https://user-images.githubusercontent.com/11618427/28630870-c0cfbb6e-722b-11e7-8216-c71d675883d2.png)
![image](https://user-images.githubusercontent.com/11618427/28630896-d6023b7e-722b-11e7-91f2-f005fe81ac46.png)

in unstable client

GrotheFAF:
I have that too, not all the time, often times directly after client start.
It seems the list is fine first and then some friends drop down in list. (when I change to Chat tab the first time after start)

Wesmania:
I did some work that ensures the chatter is sorted properly after update in my new-object-model branch. I can try porting some of this work to 0.16.0.

GrotheFAF:
another case: when I canceled a map-download while join-game from chat, the list twitched and got scrambled, some handfull of seconds later it sortet itself out (new-object-model - not cause finding?)

Wesmania:
Hmm, map downloading shouldn't affect players in any way whatsoever, so maybe it's unrelated. How did it get scrambled? Completely?

GrotheFAF:
2 groups of a few friends dropped down a few places in list. (so not all friends, about half were in the 2 groups)
and what I canceled wasn't just a map download, with that I also canceled a FA start.
and it was not reproducable (tried it many times), so could be a conicidence.

GrotheFAF:
last 'good commit': 'Add changelog entry for #663'
first (functioning) breaking commit: 'Factor off widgety part from GameItem'

GrotheFAF:
this can be closed

PR #850, opened 2017-08-25
fix friends on top

GrotheFAF:
add self.rank to reduce sorting call complexity/stack
add hack to update name to initiate sorting (which might have been the problem)
fix friends on top #813
remove unused self.status and useless self.rating

Well, kinda - just had a case where all irc were sorted in with the users on start, refaf and ok again, so something 'deeper' might be going wrong here. 
(if the initialisation doesn't go as planed: user - chatter - games, that could cause the above prbl., maybe the server update monday, will reduce or fix that prbl.)

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -132,8 +132,12 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
         else:
             self.insertTab(index, self.channels[name], name)
 
+    def sortChannels(self):
+        for channel in self.channels.values():
+            channel.sortChatters()
+
     def updateChannels(self):
-        for _, channel in self.channels.items():
+        for channel in self.channels.values():
             channel.updateChatters()
 
     def closeChannel(self, index):
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -175,8 +175,8 @@ class Chatter(QtWidgets.QTableWidgetItem):
 
     def __lt__(self, other):
         """ Comparison operator used for item list sorting """
-        firstStatus = self.getUserRank(self)
-        secondStatus = self.getUserRank(other)
+        self_rank = self.get_user_rank(self)
+        other_rank = self.get_user_rank(other)
 
         if self._me.player is not None:
             if self.user.name == self._me.player.login:
@@ -185,8 +185,8 @@ class Chatter(QtWidgets.QTableWidgetItem):
                 return False
 
         # if not same rank sort
-        if firstStatus != secondStatus:
-            return firstStatus < secondStatus
+        if self_rank != other_rank:
+            return self_rank < other_rank
 
         # Default: Alphabetical
         return self.user.name.lower() < other.user.name.lower()
@@ -200,7 +200,7 @@ class Chatter(QtWidgets.QTableWidgetItem):
         name = self.user.name
         return _id, name
 
-    def getUserRank(self, user):
+    def get_user_rank(self, user):
         # TODO: Add subdivision for admin?
         me = self._me
         _id, name = user._getIdName()
@@ -318,7 +318,7 @@ class Chatter(QtWidgets.QTableWidgetItem):
                 else:
                     self.mapItem.setIcon(icon)
 
-                self.mapItem.setToolTip(mapname)
+                self.mapItem.setToolTip(game.mapdisplayname)
             else:
                 self.mapItem.setIcon(QtGui.QIcon())
                 self.mapItem.setToolTip("")
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -790,7 +790,9 @@ class ClientWindow(FormClass, BaseClass):
 
         self.gamelogs = self.actionSaveGamelogs.isChecked()
         self.player_colors.coloredNicknames = self.actionColoredNicknames.isChecked()
-        self.friendsontop = self.actionFriendsOnTop.isChecked()
+        if self.friendsontop != self.actionFriendsOnTop.isChecked():
+            self.friendsontop = self.actionFriendsOnTop.isChecked()
+            self.chat.sortChannels()
 
         self.saveChat()
 
--- a/src/chat/channel.py
+++ b/src/chat/channel.py
@@ -87,7 +87,7 @@ class Channel(FormClass, BaseClass):
         self.name = name
         self.private = private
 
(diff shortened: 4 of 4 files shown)


Wesmania:
It's hacky, but I'm fine with it since the bug is gone in my new-object-model branch. Giving it a +2. Duke, could you merge please?

GrotheFAF:
I would consider this more a study on the problem and not a solution - cause it isn't one.

I just got another freak case letting the client run, after some time (new?) users been sorted in under irc users.

Wesmania:
Hmm, this sucks. Probably things get a bad rank after creation, when inserting - maybe try reverting that local rank variable? I think you're setting it too early in init, anyway.

GrotheFAF:
It is set again in update, and some users don't get self.rank if it isn't set in init ...even though there is the self.update() at the end of init. It is hard to understand.
And it can't be too early, because in the last case user login is long past. So to early for what? We have at that point al the ingrediants to evaluate self.rank (id, name, elevation)
At what point is a user added to players? (too much signaling for me to understand) Which is needed to get isPlayer for .rank right.

GrotheFAF:
If in chatter.set_color (irc=self.id == -1) is enough to decide, why not for rank?
So I changed it in getUserRank to: if user.id != -1: return self.RANK_USER (lose the .isPlayer())
That fixed the above mentioned case of new users being added as irc (with player color and irc logo).
(but I had **one** case where the chat userlist was all scambled up on startup. might have been a debug fluke)
and I can lose setting self.rank in init (which seems odd to be needed to begin with)

I update the commit
please someone else test this ... seems good now to me.

Wesmania:
I'm planning to switch the player list from a QListWidget to a QListView / QModel combo, the way I did with game widgets on new-object-model. This should fix all sorting / filtering woes for good. If anyone is feeling adventurous, feel free to start refactoring the chatter stuff on top of new-object-model (on a new branch preferably, since new-object-model is merge-ready right now).

We will need to make the player list an actual list, not a table, move anything not related to presentation to other classes, create appropriate delegates etc. Existing refactors of game widgets on new-object-model can serve as an inspiration.

src/client/_clientwindow.py
```diff
@@ -791,6 +791,10 @@ def updateOptions(self):
         self.gamelogs = self.actionSaveGamelogs.isChecked()
         self.player_colors.coloredNicknames = self.actionColoredNicknames.isChecked()
         self.friendsontop = self.actionFriendsOnTop.isChecked()
+        if self.friendsontop != self.actionFriendsOnTop.isChecked():
```
Wesmania: This is always false given above line.

src/chat/chatter.py
```diff
@@ -262,7 +258,9 @@ def updatePlayer(self):
             self.setIcon(util.THEME.icon("chat/countries/%s.png" % country.lower()))
             self.setToolTip(country)
 
-        self.updateAvatar()
+        if player.avatar != self.avatar:
```
Wesmania: There's no avatar member set in __init__, it probably didn't survive the rebase. It doesn't affect sorting anyway, so we can keep it how it is.

src/chat/chatter.py
```diff
@@ -175,18 +175,14 @@ def __ge__(self, other):
...
-                return False
+        if self.user.name == self._me.player.login:
+            return True
+        if other.name == self._me.player.login:
+            return False
 
         # if not same rank sort
-        if firstStatus != secondStatus:
-            return firstStatus < secondStatus
+        if self.user.rank != other.user.rank:
```
Wesmania: User has no rank, probably didn't survive rebase. I guess this is the key part of fixing sorting.

src/chat/chatter.py
```diff
@@ -175,18 +175,14 @@ def __ge__(self, other):
...
         """ Comparison operator used for item list sorting """
-        firstStatus = self.getUserRank(self)
-        secondStatus = self.getUserRank(other)
-
-        if self._me.player is not None:
-            if self.user.name == self._me.player.login:
-                return True
-            if other.user.name == self._me.player.login:
-                return False
+        if self.user.name == self._me.player.login:
```
Wesmania: _me.player can be None if the user is not logged in, so I'd keep the check.

GrotheFAF:
This is still s wreckage from your refactor. (I only rebased it to keep my branches clean)

GrotheFAF:
reduced it, cause sorting seems to work.

src/client/_clientwindow.py
```diff
@@ -790,7 +790,10 @@ def updateOptions(self):
 
         self.gamelogs = self.actionSaveGamelogs.isChecked()
         self.player_colors.coloredNicknames = self.actionColoredNicknames.isChecked()
-        self.friendsontop = self.actionFriendsOnTop.isChecked()
+        if self.friendsontop != self.actionFriendsOnTop.isChecked():
+            self.friendsontop = self.actionFriendsOnTop.isChecked()
+            self.me.relationsUpdated.emit(self.me._friends._relations)
```
Wesmania: 'relationsUpdated' should be called if player's friend/foe relation actually changes, so I don't like this bit. What we could do instead is call _sortChatters on every chat channel (by adding some invalidate_sort method to Chat and calling that).

GrotheFAF:
good hint, I still can't find my way around after the refactor...
reduced the overkill to just the right amount to get the job done.

src/chat/_chatwidget.py
```diff
@@ -132,6 +132,10 @@ def addChannel(self, name, channel, index = None):
         else:
             self.insertTab(index, self.channels[name], name)
 
+    def sortChannels(self):
+        for _, channel in self.channels.items():
```
Wesmania: `for channel in self.channels.values():`

GrotheFAF: Well, I took that line from the function 'updateChannels' right below it, why is it ok there and not here? It sure looks better and same result.

GrotheFAF: and while we're on it, should we 'unhide' _sortChatters to sortChatters? Or is there a downside?

Wesmania: That line below is some older code, and it's less elegant IMO - I think 'pick all the values' is more straightforward than 'pick all the key-value pairs, then discard the keys'. As for unhiding _sortChatters, we could do it, sure.

GrotheFAF:
done

PR #854, opened 2017-09-05
New object model

Wesmania:
Introduces the new object model for games, players and IRC users. This allows us to handle all the data related to them in one place and in organized fashion, with other client components connecting to appropriate object added / removed / updated signals.

Also, game and coop tab QListWidgets are converted to QListViews with an underlying Qt model that tracks games. This separates data from presentation completely and makes filtering and sorting more obvious and automatically updated by Qt.


Sheeo:
Mind just bumping the version to `17.0.0` rather than `0.17.0`? :) Over all looking good, though this is a rather large refactor and I'd recommend quite a lot of testing prior to release.

PR #859, opened 2017-09-08
Don't save 404 preview download

### LINKED ISSUE #852: Map downloader ignores 404 error, downloads error page as a png
Turns out that the map downloader does not check for a 404 somehow and instead downloads the generated 404 page as a PNG. If we fix it, missing PNGs will display the 'missing' icon properly rather than an empty icon.

duk3luk3:
Fixes #852

- [ ] Code has tests
- [X] Final commit includes "Fixes #issue" in commit message
- [X] Rebase onto develop
- [X] Add changelog entry

### DIFF
--- a/src/downloadManager/__init__.py
+++ b/src/downloadManager/__init__.py
@@ -53,6 +53,11 @@ class FileDownload(object):
         self._stop()
 
     def _finish(self):
+        # check status code
+        statusCode = self._dfile.attribute(QNetworkRequest.HttpStatusCodeAttribute)
+        if statusCode != 200:
+            logger.warning('Download failed: %s -> %s', self.addr, statusCode)
+            self.error = True
         self.cb_finished(self)
 
     def run(self):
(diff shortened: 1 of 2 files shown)


src/downloadManager/__init__.py
```diff
@@ -155,6 +160,7 @@ def finishedDownload(self, dler):
 
         if not dler.succeeded():
             dler.dest.remove()
+            os.unlink(filepath)
```
Wesmania, line 163: We should return at the end in this branch to avoid updating widgets with an icon that doesn't exist. I got an exception when quitting a coop game host widget.
Actually this could be a different bug, but it's still a nice thing to do.

Wesmania:
This should be perfectly functional right now, I think?

Wesmania:
@duk3luk3, I made some significant refactors to fix the download loop bug cleanly. It contains your patch, so I'll close this one on favour of #976.

PR #870, opened 2017-10-28
Add a map icon column to the chat widget

aeoncleanse:
Rebased and adapted from #331.
This adds a new column which shows a map preview in the player list of the chat tab. There is a settings menu toggle for it, and the chat resizes itself to fit

- [ ] Code has tests
- [ ] Add changelog entry

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -297,6 +297,7 @@ class Chatter(QtWidgets.QTableWidgetItem):
 
     def updateGame(self):
         # Status icon handling
+
         game = self.user_game
         player = self.user_player
         if game is not None and not game.closed():
@@ -307,6 +308,9 @@ class Chatter(QtWidgets.QTableWidgetItem):
             elif game.state == GameState.PLAYING:
                 self.statusItem.setIcon(util.THEME.icon("chat/status/playing.png"))
                 self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+
+            # We're in game, show the map if toggled on
+            if util.settings.value("chat/chatmaps", False):
                 mapname = game.mapname
                 icon = maps.preview(mapname)
                 if not icon:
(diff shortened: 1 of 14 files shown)


src/chat/chatter.py
```diff
@@ -298,6 +307,14 @@ def updateGame(self):
             elif game.state == GameState.PLAYING:
                 self.statusItem.setIcon(util.THEME.icon("chat/status/playing.png"))
                 self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
+                mapname = game.mapname
+                icon = maps.preview(mapname)
+                if not icon:
+                    self.chat_widget.client.downloader.downloadMapPreview(mapname, self)
```
Wesmania: The object you pass here will have a setIcon method called on it once the download is complete. You want to pass mapItem here, else the map will show in place of the flag.

GrotheFAF:
map doesn't go away when game or lobby ends

aeoncleanse:
Do we need a new hook into the refresh function from a lobby end trigger? Feel free to amend the PR or whatever if you know how to fix this

GrotheFAF:
The code doesn't cover the case from user in game to user not in game.
I have a fix, will come as part of my next PR.

Wesmania:
The Player class from the model folder has a newCurrentGame signal that's called with the current and and previous game object (any of them can be none if the player wasn't or no longer is playing a game).

GrotheFAF:
Here it's not a signaling problem, just a case not covered.
But I have a signaling problem: I need gameset.newLiveReplay as signal in chatter ...

Wesmania:
Hmm, this info should be possible to receive via Game object, either through update or a signal. Check there, if there is neither then we'll have to add one.

GrotheFAF:
I tried, but I'm out of ideas how to do that.

Wesmania:
Chatter already has user_game getters and setters. We can connect to and disconnect from game's liveReplayAvailable signal there, the way it's done with the other signals. Then it's just a matter of handling this signal.
If you feel it's getting too complex, we could delegate keeping track of the map icon state to some class.

GrotheFAF:
(this is not about the map icon - it's about the status icon -> using liveReplayAvailable)
well, with user_game setter it kinda works, but some games (hence there chatters) slip through, and I had to add a second here and there to make it work at all.

GrotheFAF:
guess I will split the map icon fix and status icon spice into 2 PRs

Issue #762, opened 2017-06-27
From chat user 'View Replays in Vault' doesn't show replays if rating <0


GrotheFAF:
I searched User Yucassha and it only showed one replay. (and not the one I just played against him)
After I set 'Min rating' to -100 the other replays (ladder rating below zero) showed up.

I would expect that chat user 'View Replays in Vault' shows all replays of a user.

One solution would be to set minRating  in chatter.py/viewVaultReplay to a reasonable negative value.

A better solution would be a flag that makes the server ignores the minRating in the query. (which could actually be if minRating == 0)

Wesmania:
Somebody could legitimately want to set minRating to 0. We could just send a really low negative value if the restriction is "disabled".

ghost:
This issue should be closed by the way (cf PR)

Wesmania:
We'll close it once it's merged. You can add a 'Fixes #762' line to the commit description to autoclose it at merge.

ghost:
Ugh, I may be wrong but pretty sure the PR has already been merged : https://github.com/FAForever/client/pull/878 so...?

I see for the autoclose, I didn't know that. I did 'Fix issue #762' in the commit message instead of 'Fixes #762'.

Wesmania:
Alright.

PR #878, opened 2017-11-13
#762 - fix - From chat user 'View Replays in Vault' doesn't show replays if rating <0 

ghost:
Subject : 
"I searched User Yucassha and it only showed one replay. (and not the one I just played against him)
After I set 'Min rating' to -100 the other replays (ladder rating below zero) showed up.
I would expect that chat user 'View Replays in Vault' shows all replays of a user."
Now when you start the client and look for a replay, it will have by default -2000, so all the replays will show when you search for an user.

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry


GrotheFAF:
This will not fix the subject problem, because of https://github.com/FAForever/client/blob/develop/src/chat/chatter.py#L476 starts searchVault with minrating=0.
(and isn't -2000 a bit much? only the bottom 100 in ladder rating go below -200)

ghost:
Ugh, the replay vault can be based either on ladder rating if ladder1v1 or global rating otherwise. -2000 may be a bit too much but I don't know what is the minimal global/ladder rated player on faf so...

And I tried with "Yucassha" and it worked fine, I guess because the default value is now -2000 instead of 0 with this change. And for your line not really sure about it (was confused because you can precise minrating when you search replays, so it is not locked to 0)

GrotheFAF:
You tried Yucassha" from Chat Lobby, right click user -> 'View Replay in Vault' ? (see subject) 
... which leads to the line.

ghost:
Ahh ! Ok I didn't think of this case, good catch.

I searched Yucassha in the replay vault directly (which solved it) but right click on his name no I haven't tried it (and I guess for this one it is not solved).

ghost:
Ok I fixed this case too when you look for the replays of the player with right click too, I lowered to -1400.
I read someone had -800 so I'm not that far from the truth. It should be fine now.

Issue #875, opened 2017-11-12
Divide game.log to multiples files with replayid 


ghost:
In addition to game.log have multiple files having this name : 
game.log.REPLAYIDOFTHEGAME.log

=>Be sure to limit their number (max 10)
=>Be sure that game.log can't exceed the game.log limit

PR #881, opened 2017-11-13
Divide game.log to different files with replayid (Feature #875)

ghost:
Divide game.log to different files with replayid. For example : 
game.log.Replayid7050168.log
game.log.Replayid7050222.log
So you can now check the game log related to the replayid instead having a single file. The game.log is still kept and renamed to game.log.full.log.
Also added a limit to size for the game.log file.
It needs some more testing though to be sure it is working and is really correct (that the game log files are really only about the replay id game)

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [ ] Add changelog entry

### DIFF
--- a/src/fa/game_session.py
+++ b/src/fa/game_session.py
@@ -39,7 +39,7 @@ class GameSession(QObject):
         self.game_visibility = None
         self.game_map = None
         self.game_password = None
-        self.game_has_started = 0
+        self.game_has_started = False
 
         # Subscribe to messages targeted at 'game' from the server
         client.lobby_dispatch.subscribe_to('game', self.handle_message)
@@ -170,7 +170,7 @@ class GameSession(QObject):
                 # TODO: Eagerly initialize the game by hosting/joining early
                 pass
             elif args[0] == 'Launching':
-                self.game_has_started = 1
+                self.game_has_started = True
         elif command == 'Rehost':
             self._rehost = True
         elif command == 'GameFull':
@@ -186,28 +186,27 @@ class GameSession(QObject):
         self._add_to_full_gamelog()
 
     def _limit_replayid_gamelogs(self):
-        nb_replayid_files = 0
         files = os.listdir(util.LOG_DIR)
         files = map(lambda f: os.path.join(util.LOG_DIR, f), files)
-        files = sorted(files, key = os.path.getmtime)
+        files = sorted(files, key=os.path.getmtime)
         files = list(map(os.path.basename, files))
         replay_files = [e for e in files if "Replayid" in e]
         while len(replay_files) >= util.MAX_NUMBER_REPLAYID_LOG_FILE:
-            os.remove(os.path.join(util.LOG_DIR,replay_files[0]))
+            os.remove(os.path.join(util.LOG_DIR, replay_files[0]))
             replay_files.pop(0)
 
     def _add_to_full_gamelog(self):
         if os.path.isfile(util.LOG_FILE_GAME):
-            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.full.log', 'a+') as dst: dst.write(src.read())
+            with open(util.LOG_FILE_GAME, 'r') as src, open(os.path.join(util.LOG_DIR, 'game.log'), 'a+') as dst: dst.write(src.read())
 
     def _clear_initial_gamelog(self):
         with open(util.LOG_FILE_GAME, 'w') as src : pass
 
     def _write_new_replayid_gamelogs(self, replay_ID):
-        if self.game_has_started == 1:
+        if self.game_has_started:
             if os.path.isfile(util.LOG_FILE_GAME):
-                with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.Replayid'
-                + str(replay_ID) + '.log', 'w+') as dst: dst.write(src.read())
+                end_logname = "{}{}".format(str(replay_ID), '.log')
+                with open(util.LOG_FILE_GAME, 'r') as src, open("{}{}".format(os.path.join(util.LOG_DIR, 'Replayid'), end_logname), 'w+') as dst: dst.write(src.read())
             else:
                 self._logger.error('No game log found to create the replayid game log file')
 
@@ -219,7 +218,7 @@ class GameSession(QObject):
 
         self._limit_replayid_gamelogs()
         self._write_new_replayid_gamelogs(self.game_uid)
-        self.game_has_started = 0
+        self.game_has_started = False
         self._add_to_full_gamelog()
         self._clear_initial_gamelog()
 
--- a/src/fa/game_session.py
+++ b/src/fa/game_session.py
@@ -7,6 +7,9 @@ from decorators import with_logger
 from fa.game_connection import GPGNetConnection
 from fa.game_process import instance
 
+from fa.game_process import GameProcess, instance as game_process_instance
+import util
+import os
 
 class GameSessionState(IntEnum):
     # Game services are entirely off
@@ -36,6 +39,7 @@ class GameSession(QObject):
         self.game_visibility = None
         self.game_map = None
         self.game_password = None
+        self.game_has_started = 0
 
         # Subscribe to messages targeted at 'game' from the server
         client.lobby_dispatch.subscribe_to('game', self.handle_message)
@@ -165,6 +169,8 @@ class GameSession(QObject):
             elif args[0] == 'Lobby':
                 # TODO: Eagerly initialize the game by hosting/joining early
                 pass
+            elif args[0] == 'Launching':
+                self.game_has_started = 1
         elif command == 'Rehost':
             self._rehost = True
         elif command == 'GameFull':
@@ -177,12 +183,45 @@ class GameSession(QObject):
 
     def _launched(self):
         self._logger.info("Game has started")
+        self._add_to_full_gamelog()
+
+    def _limit_replayid_gamelogs(self):
+        nb_replayid_files = 0
+        files = os.listdir(util.LOG_DIR)
+        files = map(lambda f: os.path.join(util.LOG_DIR, f), files)
+        files = sorted(files, key = os.path.getmtime)
+        files = list(map(os.path.basename, files))
+        replay_files = [e for e in files if "Replayid" in e]
+        while len(replay_files) >= util.MAX_NUMBER_REPLAYID_LOG_FILE:
+            os.remove(os.path.join(util.LOG_DIR,replay_files[0]))
+            replay_files.pop(0)
+
+    def _add_to_full_gamelog(self):
+        if os.path.isfile(util.LOG_FILE_GAME):
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.full.log', 'a+') as dst: dst.write(src.read())
+
+    def _clear_initial_gamelog(self):
+        with open(util.LOG_FILE_GAME, 'w') as src : pass
+
+    def _write_new_replayid_gamelogs(self, replay_ID):
+        if self.game_has_started == 1:
+            if os.path.isfile(util.LOG_FILE_GAME):
+                with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.Replayid'
+                + str(replay_ID) + '.log', 'w+') as dst: dst.write(src.read())
+            else:
+                self._logger.error('No game log found to create the replayid game log file')
 
     def _exited(self, status):
         self._game_connection = None
         self.state = GameSessionState.OFF
         self._logger.info("Game has exited with status code: {}".format(status))
         self.send('GameState', ['Ended'])
+
+        self._limit_replayid_gamelogs()
+        self._write_new_replayid_gamelogs(self.game_uid)
+        self.game_has_started = 0
+        self._add_to_full_gamelog()
+        self._clear_initial_gamelog()
 
         if self._rehost:
             self._client.host_game(title=self.game_name,
--- a/src/util/__init__.py
+++ b/src/util/__init__.py
@@ -57,7 +57,7 @@ REPLAY_DIR = os.path.join(APPDATA_DIR, "replays")
 # This contains all Lobby, Chat and Game logs
 LOG_DIR = os.path.join(APPDATA_DIR, "logs")
 LOG_FILE_FAF = os.path.join(LOG_DIR, 'forever.log')
-LOG_FILE_GAME = os.path.join(LOG_DIR, 'game.log')
+LOG_FILE_GAME = os.path.join(LOG_DIR, 'temporary.log')
 LOG_FILE_REPLAY = os.path.join(LOG_DIR, 'replay.log')
 MAX_NUMBER_REPLAYID_LOG_FILE = 10
 
@@ -154,9 +154,9 @@ try:
     if os.path.isfile(LOG_FILE_FAF):
         if os.path.getsize(LOG_FILE_FAF) > LOGFILE_MAX_SIZE:
             os.remove(LOG_FILE_FAF)
-    if os.path.isfile(LOG_FILE_GAME + '.full.log'):
-        if os.path.getsize(LOG_FILE_GAME + '.full.log') > LOGFILE_MAX_SIZE:
-            os.remove(LOG_FILE_GAME + '.full.log')
+    if os.path.isfile(os.path.join(LOG_DIR, "game.log")):
+        if os.path.getsize(os.path.join(LOG_DIR, "game.log")) > LOGFILE_MAX_SIZE:
+            os.remove(os.path.join(LOG_DIR, "game.log"))
 except:
     pass
 
--- a/src/util/__init__.py
+++ b/src/util/__init__.py
@@ -59,6 +59,7 @@ LOG_DIR = os.path.join(APPDATA_DIR, "logs")
 LOG_FILE_FAF = os.path.join(LOG_DIR, 'forever.log')
 LOG_FILE_GAME = os.path.join(LOG_DIR, 'game.log')
 LOG_FILE_REPLAY = os.path.join(LOG_DIR, 'replay.log')
+MAX_NUMBER_REPLAYID_LOG_FILE = 10
 
 # This contains the game binaries (old binFAF folder) and the game mods (.faf files)
 BIN_DIR = os.path.join(APPDATA_DIR, "bin")
@@ -153,6 +154,9 @@ try:
     if os.path.isfile(LOG_FILE_FAF):
         if os.path.getsize(LOG_FILE_FAF) > LOGFILE_MAX_SIZE:
             os.remove(LOG_FILE_FAF)
+    if os.path.isfile(LOG_FILE_GAME + '.full.log'):
+        if os.path.getsize(LOG_FILE_GAME + '.full.log') > LOGFILE_MAX_SIZE:
+            os.remove(LOG_FILE_GAME + '.full.log')
 except:
     pass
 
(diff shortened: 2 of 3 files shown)


src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
         self.state = GameSessionState.OFF
         self._logger.info("Game has exited with status code: {}".format(status))
         self.send('GameState', ['Ended'])
 
+        replayID=''
+        with open(util.LOG_FILE_GAME,'r') as inF:
+            for line in inF:
+                if 'gpgnet://localhost/' in line:
+                    try:
+                        replayID=line.split('localhost/')[1].split('/')[0]
```
Wesmania: The class already has a game_uid member that gets set at game launch. Let's use it instead of hacking things.

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+        try:
+            nb_replayid_files=0
+            list_game_replayid_files=os.listdir(util.LOG_DIR)
+            size_replayid_files=len(list_game_replayid_files)-1
+            while size_replayid_files >= 0:
+                if "Replayid" not in list_game_replayid_files[size_replayid_files]:
+                    list_game_replayid_files.remove(list_game_replayid_files[size_replayid_files])
+                else:
+                    nb_replayid_files+=1
+                size_replayid_files-=1
```
Wesmania: This whole loop is really unpythonic. I'd rather see some list comprehension here to filter replay log files.

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+            if nb_replayid_files >=11:
```
Wesmania: This looks really arbitrary. Why 11 and why remove a single replay? If we want to rotate replay files, then let's do it where the main logfile is rotated. Also note that listdir returns entries in arbitrary order, so we'd have to make sure we order them by modification time or replay id before deleting the logs.

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+                os.remove(util.LOG_DIR+"\\"+list_game_replayid_files[0])
```
Wesmania: Use os.path.join instead of a hardcoded backslash, it won't work on unix.

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.Replayid'+str(replayID)+'.log', 'w+') as dst: dst.write(src.read())
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.full.log', 'a+') as dst: dst.write(src.read())
```
Wesmania: These lines repeat a couple of times, maybe make them a method, something like copy_log_to_file(self, dst)?

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+            with open(util.LOG_FILE_GAME, 'w') as src : pass
```
Wesmania: Why would this be necessary?

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+        except :
```
Wesmania: Catch-all handlers are horrible for debugging, please figure out and catch exactly the exceptions you expect.

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
...
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.full.log', 'a+') as dst: dst.write(src.read())
```
Wesmania: Duplicated above line. Maybe let's just do it unconditionally outside of try/catch?

src/fa/game_session.py
```diff
@@ -177,13 +178,41 @@ def _turn_state_changed(self, val):
 
     def _launched(self):
         self._logger.info("Game has started")
+        with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.full.log', 'a+') as dst: dst.write(src.read())
```
Wesmania: I'm not sure if you shouldn't clear the LOG_FILE_GAME here. It's hard to tell when the code isn't split up into functions.

Wesmania (changes requested):
There's some stuff I'd like to see fixed. Most of all, please move parts of the code to smaller methods with descriptive names - it'd make it much easier to verify that the code works as it should.

src/fa/game_session.py
ghost: Oh indeed, didn't notice there was game_uid there already made. Removed this hacky solution

src/fa/game_session.py
ghost: fixed

src/fa/game_session.py
ghost: Arbitrary indeed. To avoid having 200 log files for example, just keep the most recent ones, it can be 11 or 10 or something else (whatever number you think is fine).

And it removes only one because the limitation is called every time you end a game. (so at most you have +1 replay before the next time it is called)

ghost:
WIP, not finished. I will tell once you can look onto it again, and indeed I should have methods earlier, it is much cleaner.

TODO : 
>clean/verify code
>transform the loop into a more pythonic one 
>listdir result have to be ordered (to verify, because I think the result is already ordered by name, and the most ancient one will be the first one, since replayid increase every time you play)
>Test max log file limit
>Test the game log system

Wesmania:
You can use 'git rebase -i' to squash these commits together into one, it's really convenient too.

ghost:
Yes I like rebase, I've been using for example  'git rebase -i HEAD~2' . I don't have any more issue for the time being with git.

src/fa/game_session.py
ghost: code split up into function

ghost: code split up into function

src/fa/game_session.py
ghost: Transformed it into a more pythonic one, I think, maybe

ghost:
Ok what about now ? Thanks for the suggestions btw.

Don't hesitate if you see anything else to do or something you don't understand.

For game_has_started variable : 
I had to add a variable to know when the game is really starting 'game_has_started' because '_launched' method means that the forgedalliance game started so they are in lobby but not that the game really started (=> no replay id file should be made in this case, people can host a game and then leave it without it starting).
So I had to find a way to know exactly when the game really starts (did it by catching the launching message). 

Why clear the file when you join/host a game ?
Because the replayid file has to be only for the potentiel game concerned, no other things. And it is wrong to assume than the log should be cleared only after a game has finished , example : 
-> game finished  (log is cleared)
-> join a game called n1 (log has info)
-> leave the game (log still has info)
-> join another game called n2 (log has more info => PROBLEM : the log is concerning another game too, n1 and n2)
So that is why you clear the log when you join/host a game.

src/fa/game_session.py
```diff
@@ -165,6 +173,8 @@ def _on_game_message(self, command, args):
             elif args[0] == 'Lobby':
                 # TODO: Eagerly initialize the game by hosting/joining early
                 pass
+            elif args[0] == 'Launching':
```
Wesmania, line 172: If you're doing it here, I guess update_if_game_has_launched is not necessary.

src/fa/game_session.py
```diff
@@ -177,13 +187,45 @@ def _turn_state_changed(self, val):
 
     def _launched(self):
         self._logger.info("Game has started")
+        self._add_to_full_gamelog()
+
+    def _limit_replayid_gamelogs(self):
+        nb_replayid_files=0
+        list_game_replayid_files=os.listdir(util.LOG_DIR)
+        list_game_replayid_files.sort()
+        for element in list_game_replayid_files[::1]:
```
Wesmania: How about `replay_files = [e for e in list_game_replayid_files if "Replayid" in e]`?

src/fa/game_session.py
```diff
@@ -177,13 +187,45 @@ def _turn_state_changed(self, val):
 
     def _launched(self):
         self._logger.info("Game has started")
+        self._add_to_full_gamelog()
+
+    def _limit_replayid_gamelogs(self):
+        nb_replayid_files=0
+        list_game_replayid_files=os.listdir(util.LOG_DIR)
+        list_game_replayid_files.sort()
```
Wesmania: This is not an ideal solution, since we could (in theory) play a game with a later uid, then one with an earlier one (a short ladder game, then a custom game, for example). We could sort by modification time like:
```
files = os.listdir(util.LOG_DIR)
files = map(lambda f: os.path.join(util.LOG_DIR, f), files)
files= sorted(files, key=os.path.getmtime)
files = list(map(os.path.basename, files))
```

src/fa/game_session.py
```diff
@@ -177,13 +187,45 @@ def _turn_state_changed(self, val):
...
+        for element in list_game_replayid_files[::1]:
+            if "Replayid" not in element:
+                list_game_replayid_files.remove(element)
+            else:
+                nb_replayid_files+=1
+        if nb_replayid_files >=util.MAX_NUMBER_REPLAYID_LOG_FILE:
```
Wesmania: We could as well use len(list_game_replayid_files) and remove files until we get below the threshold, just to be sure.

src/fa/game_session.py
```diff
@@ -177,13 +187,45 @@ def _turn_state_changed(self, val):
...
+    def _clear_initial_gamelog(self):
+        with open(util.LOG_FILE_GAME, 'w') as src : pass
+
+    def _write_new_replayid_gamelogs(self,replay_ID):
+        if self.game_has_started == 1:
+            if os.path.isfile(util.LOG_FILE_GAME):
+                with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.Replayid'+str(replay_ID)+'.log', 'w+') as dst: dst.write(src.read())
+            else:
+                self._logger.error('No game log found to create the replayid game log file')
+            self.game_has_started=0
```
Wesmania: I'd put this line directly in _exited.

src/fa/game_session.py
ghost: Mmh, I am not 100% sure because from what I saw there is '_on_game_message' and 'send' methods. I was thinking the 'launching' can either be when you start a hosted game or when the game is started by someone else (so you receive or send kinda...) but maybe I'm totally wrong.

I haven't tried without removing the 'update_if_game_has_launched'. Will try.

src/fa/game_session.py
ghost: Sure

src/fa/game_session.py
ghost: Oh, I didn't know of this case of ladder/custom game for uid. Thanks fixed.

src/fa/game_session.py
ghost: Ok moved it to _exited. fixed

src/fa/game_session.py
ghost: Ok, done

src/fa/game_session.py
ghost: Ok I removed the 'update_if_game_has_launched' method and it still worked.

ghost:
Ok did all the suggestion, as usual, any question/suggestion/something or something else don't hesitate

Wesmania:
Some merging mess happened when rebasing, you ended up merging the develop branch into your PR. Try resetting your branch back to 2c62bac27 (original PR), then doing 'git pull --rebase upstream develop' and fixing merge conflicts as git instructs. Fixing conflicts at rebase is pretty confusing, I admit :/

ghost:
Yeah... Hum. I'm not surprised at all of your answer when I noticed these 2 commits which were not usual at all, it happened because I was having a conflict and... idk .
Will write again once it's fixed.

ghost:
Ok, should be fixed now

src/fa/game_session.py
```diff
@@ -36,6 +39,7 @@ def __init__(self, client, connectivity):
         self.game_visibility = None
         self.game_map = None
         self.game_password = None
+        self.game_has_started=0
```
Wesmania: Whitespace around '='

src/fa/game_session.py
```diff
@@ -165,6 +169,8 @@ def _on_game_message(self, command, args):
             elif args[0] == 'Lobby':
                 # TODO: Eagerly initialize the game by hosting/joining early
                 pass
+            elif args[0] == 'Launching':
+                self.game_has_started=1
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
 
     def _launched(self):
         self._logger.info("Game has started")
+        self._add_to_full_gamelog()
+
+    def _limit_replayid_gamelogs(self):
+        nb_replayid_files=0
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
...
+        files = os.listdir(util.LOG_DIR)
+        files = map(lambda f: os.path.join(util.LOG_DIR, f), files)
+        files= sorted(files, key=os.path.getmtime)
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
...
+        files = list(map(os.path.basename, files))
+        replay_files = [e for e in files if "Replayid" in e]
+        while len(replay_files) >=util.MAX_NUMBER_REPLAYID_LOG_FILE:
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
...
+            replay_files.pop(0)
+
+    def _add_to_full_gamelog(self):
+        if os.path.isfile(util.LOG_FILE_GAME):
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.full.log', 'a+') as dst: dst.write(src.read())
+
+    def _clear_initial_gamelog(self):
+        with open(util.LOG_FILE_GAME, 'w') as src : pass
+
+    def _write_new_replayid_gamelogs(self,replay_ID):
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
...
+        if self.game_has_started == 1:
+            if os.path.isfile(util.LOG_FILE_GAME):
+                with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.Replayid'+str(replay_ID)+'.log', 'w+') as dst: dst.write(src.read())
```
Wesmania: Put it in 2 lines.

src/fa/game_session.py
```diff
@@ -177,13 +183,45 @@ def _turn_state_changed(self, val):
...
+        files= sorted(files, key=os.path.getmtime)
+        files = list(map(os.path.basename, files))
+        replay_files = [e for e in files if "Replayid" in e]
+        while len(replay_files) >=util.MAX_NUMBER_REPLAYID_LOG_FILE:
+            os.remove(os.path.join(util.LOG_DIR,replay_files[0]))
+            replay_files.pop(0)
+
+    def _add_to_full_gamelog(self):
+        if os.path.isfile(util.LOG_FILE_GAME):
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME+'.full.log', 'a+') as dst: dst.write(src.read())
```
Wesmania: spaces around '+'

src/util/__init__.py
```diff
@@ -153,6 +154,9 @@ def developer():
     if os.path.isfile(LOG_FILE_FAF):
         if os.path.getsize(LOG_FILE_FAF) > LOGFILE_MAX_SIZE:
             os.remove(LOG_FILE_FAF)
+    if os.path.isfile(LOG_FILE_GAME+'.full.log'):
```
Wesmania: ws

Wesmania (changes requested):
Formatting stuff, I suggest integrating a linter like autopep8 into the editor / IDE for convenience.

ghost:
ok done all the ws fix I think.

Wesmania:
Alright, I'll give it a round of testing tomorrow.

Wesmania (approved):

Wesmania:
One things I noticed when testing - generated log names are very ugly, in likes of 'game.log.full.log'. I think it would be better to explicitly call game.log the temporary log, rename 'full.log' to 'game.log' and remove the 'game.log' prefix from game ID logs.

src/fa/game_session.py
```diff
@@ -36,6 +39,7 @@ def __init__(self, client, connectivity):
         self.game_visibility = None
         self.game_map = None
         self.game_password = None
+        self.game_has_started = 0
```
Wesmania: Use a boolean here.

src/fa/game_session.py
```diff
@@ -177,13 +183,46 @@ def _turn_state_changed(self, val):
 
     def _launched(self):
         self._logger.info("Game has started")
+        self._add_to_full_gamelog()
+
+    def _limit_replayid_gamelogs(self):
+        nb_replayid_files = 0
```
Wesmania: unused

src/fa/game_session.py
```diff
@@ -177,13 +183,46 @@ def _turn_state_changed(self, val):
...
+        files = os.listdir(util.LOG_DIR)
+        files = map(lambda f: os.path.join(util.LOG_DIR, f), files)
+        files = sorted(files, key = os.path.getmtime)
```
Wesmania: Either I was wrong here or didn't catch this one - no whitespace around keyword parameter assignment

src/fa/game_session.py
```diff
@@ -177,13 +183,46 @@ def _turn_state_changed(self, val):
...
+        files = list(map(os.path.basename, files))
+        replay_files = [e for e in files if "Replayid" in e]
+        while len(replay_files) >= util.MAX_NUMBER_REPLAYID_LOG_FILE:
+            os.remove(os.path.join(util.LOG_DIR,replay_files[0]))
```
Wesmania: ws

src/fa/game_session.py
```diff
@@ -177,13 +183,46 @@ def _turn_state_changed(self, val):
...
+            replay_files.pop(0)
+
+    def _add_to_full_gamelog(self):
+        if os.path.isfile(util.LOG_FILE_GAME):
+            with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.full.log', 'a+') as dst: dst.write(src.read())
+
+    def _clear_initial_gamelog(self):
+        with open(util.LOG_FILE_GAME, 'w') as src : pass
```
Wesmania: If we explicitly make it a temporary file, we can just remove it here.

src/fa/game_session.py
```diff
@@ -177,13 +183,46 @@ def _turn_state_changed(self, val):
...
+
+    def _write_new_replayid_gamelogs(self, replay_ID):
+        if self.game_has_started == 1:
+            if os.path.isfile(util.LOG_FILE_GAME):
+                with open(util.LOG_FILE_GAME, 'r') as src, open(util.LOG_FILE_GAME + '.Replayid'
```
Wesmania: Use format() here instead of concatenating strings.

src/fa/game_session.py
ghost: You mean that once I renamed all the files as you said, I can just remove this function ?

Wesmania:
I'll grab this one as my own. I'll do some work on it, test it and hopefully merge it tomorrow.

Wesmania:
I decided to take an approach of giving the log file as an argument directly. It does mean that lobby logs are also separate, but I think it's a preferable approach. I'll close this and reopen a new one.

PR #891, opened 2017-11-28
fix #890 chatter map icon

### LINKED ISSUE #890: fix chatter map case miss
the map doesn't go away when game or lobby ends

GrotheFAF:
move map handling outside of status handling for clarity
fix case after game or lobby end still showing map icon

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -297,7 +297,6 @@ class Chatter(QtWidgets.QTableWidgetItem):
 
     def updateGame(self):
         # Status icon handling
-
         game = self.user_game
         player = self.user_player
         if game is not None and not game.closed():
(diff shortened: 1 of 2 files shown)


src/chat/chatter.py
```diff
@@ -308,24 +308,24 @@ def updateGame(self):
...
-
-                self.mapItem.setToolTip(game.mapdisplayname)
-            else:
-                self.mapItem.setIcon(QtGui.QIcon())
-                self.mapItem.setToolTip("")
         else:
             self.statusItem.setIcon(QtGui.QIcon())
             self.statusItem.setToolTip("Idle")
 
+        # Map icon handling - if we're in game, show the map if toggled on
```
Wesmania: I like this. If we already pulled it out, maybe we can put it in a separate method altogether, like update_game_map_icon(self, game).

GrotheFAF: That might make sense if we had a extra route for map change, else it's closely related to game.state. The signaling for seperate methods would be ... complicated.

(while we're here: Does self._verifySortOrder() makes any sense here in updateGame? Nothing related to sorting with state & map icon.

Wesmania: We don't need to make a separate control flow for the method, we can just call it from here. It'd be just a code organization matter, to keep functions small and focused on one thing.
_verifySortOrder is just a dirty fix to ensure that chatters are well-sorted after some updates (downside of using qt widget lists instead of qt models like games are done now). Map updates don't affect the order, so we don't need to call it here.

GrotheFAF:
like this? I don't see the advantage.

PR #892, opened 2017-11-30
add host & live-game-delay to user status in Chat

GrotheFAF:
updatePlayer:
refactor tooltip_str, icon_str
add __.png for no country2ip
remove set_color and verifySortOrder (updateUser is doing that just fine)

updateGame:
refactor gamestr, tooltip_str, icon_str
use other single sword (gold it) for host of hosted game
add title, mod, map and players info to user status tooltip (and remove url)
add playing5.png for live-game delay status icon
add liveReplayAvailable connection for updateGame
add unknown.png for user with unknown game.state (you never know...)

fix QUrl for pyqt5


GrotheFAF:
The wizzards of signaling might check if I did it right.

src/chat/chatter.py
```diff
@@ -141,6 +141,8 @@ def user_game(self, value):
 
         if self._user_game is not None:
             self._user_game.gameUpdated.connect(self.updateGame)
+            if self._user_game.state == GameState.PLAYING:
+                self._user_game.liveReplayAvailable.connect(self.updateGame)
```
Wesmania: user_game is only set when a player's game changes, so if we join a game that's in another state (like OPEN), this will never get connected. We should always connect to it and check any extra conditions in the handler. Also make sure we disconnect from this signal when we change user_game again.

src/chat/chatter.py
```diff
@@ -248,19 +250,16 @@ def updateUser(self):
...
             self.rankItem.setIcon(util.THEME.icon("chat/rank/civilian.png"))
             self.rankItem.setToolTip("IRC User")
             return
 
-        country = player.country
-        if country is not None:
-            self.setIcon(util.THEME.icon("chat/countries/%s.png" % country.lower()))
-            self.setToolTip(country)
+        country = player.country  # server sends '' for no ip2country-resolution
+        self.setIcon(util.THEME.icon("chat/countries/%s.png" % (country or '__').lower()))
```
Wesmania: Changed logic, now we're checking truthiness of 'country' instead of comparing it to None. If it can be an empty string, it'd be better to check explicitly.

src/chat/chatter.py
```diff
@@ -273,43 +272,56 @@ def updatePlayer(self):
...
+        self.rankItem.setIcon(util.THEME.icon("chat/rank/%s.png" % icon_str))
+        self.rankItem.setToolTip(tooltip_str)
 
     def updateGame(self):
         # Status icon handling
         game = self.user_game
-        player = self.user_player
         if game is not None and not game.closed():
-            url = game.url(player.id)
+            game_str = "<br/>title: " + game.title + "<br/>mod: " + game.featured_mod + "<br/>map: " + \
```
Wesmania: Please use Python's format() function for this kind of stuff. This string formatting is complex enough that it deserves its own method IMO.

src/chat/chatter.py
```diff
@@ -273,43 +272,56 @@ def updatePlayer(self):
...
+                else:
+                    icon_str, tooltip_str = "lobby", "<b>In"
             elif game.state == GameState.PLAYING:
-                self.statusItem.setIcon(util.THEME.icon("chat/status/playing.png"))
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
-        else:
-            self.statusItem.setIcon(QtGui.QIcon())
-            self.statusItem.setToolTip("Idle")
+                if game.launched_at:
+                    delay = time.time() - game.launched_at + 1  # + 1 async jitter
```
Wesmania: We shouldn't do it like this. If we want to know precisely if the game has a live replay already, we can add extra data to Game class that gets set before it emits the live replay signal and expose it as a property.

Wesmania (changes requested):

src/chat/chatter.py
GrotheFAF: hmm ... isn't joining a game, when it's OPEN, the usual case?
I only need that connection, if the game starts and that is a game.state change (we see chatter status change if a game starts .. so updateGame is called) ...?
So you suggest I lose the line 144?
I did a disconnect (like gameUpdated a few lines up), but that got me a lot of 'not connected' bug messages (mentioning the connect line, which step debug fine, so I don't know how to debug that and find out what's wrong. ... I guess delayed action on a connection makes it tricky)

src/chat/chatter.py
GrotheFAF: The None case does not happen here, but the '' does.
I thought you like it compact. You want some if ... else ... ?

src/chat/chatter.py
GrotheFAF: well, it grew slowly over time ... it's not that complex .. but I know you prefer some formatting ;-)
I'll see what I can do ....

src/chat/chatter.py
GrotheFAF: oh, you're so right ... "game.has_live_replay"    was already there, how did I miss that ...

src/chat/chatter.py
GrotheFAF: losing line 144, and putting the disconnect line back in done the trick. happy signaling ;-)

src/chat/chatter.py
Wesmania: We can do something like `country = player.country if player.country != '' else '__'`. It's not perfect, but Python doesn't have a Maybe monad just yet :/

src/chat/chatter.py
```diff
@@ -135,12 +135,16 @@ def user_game(self):
...
         if self._user_game is not None:
             self._user_game.gameUpdated.disconnect(self.updateGame)
+            self._user_game.liveReplayAvailable.disconnect(self.updateGame)
 
         self._user_game = value
         self.updateGame()
 
         if self._user_game is not None:
             self._user_game.gameUpdated.connect(self.updateGame)
+            #game = self._user_game
```
Wesmania: Don't leave commented-out code.

src/chat/chatter.py
```diff
@@ -273,43 +274,52 @@ def updatePlayer(self):
...
         if game is not None and not game.closed():
-            url = game.url(player.id)
+            formatting = "<br/>title: {}<br/>mod: {}<br/>map: {}<br/>players: {} / {}<br/>id: {}"
+            game_str = formatting.format(game.title, game.featured_mod, maps.getDisplayName(game.mapname),
+                                         str(game.num_players), str(game.max_players), str(game.uid))
             if game.state == GameState.OPEN:
-                self.statusItem.setIcon(util.THEME.icon("chat/status/lobby.png"))
-                self.statusItem.setToolTip("In Game Lobby<br/>"+url.toString())
+                if game.password_protected:
+                    game_str = " (private) Game Lobby:</b>" + game_str
```
Wesmania: You either missed a \<b\>, or it should be a \</br\>.

src/chat/chatter.py
```diff
@@ -273,43 +274,52 @@ def updatePlayer(self):
...
+                else:
+                    icon_str, tooltip_str = "playing5", "<b>Playing:</b>" + "   LIVE DELAY (5 Min)"
+            else:  # game.state == something else
+                icon_str, tooltip_str = "unknown", "<b>Playing maybe ...</b>"
+
+        else:  # game is None or closed
+            icon_str, tooltip_str, game_str = "none", "Idle", ""
+
+        self.statusItem.setIcon(util.THEME.icon("chat/status/%s.png" % icon_str))
+        self.statusItem.setToolTip(tooltip_str + game_str)
```
Wesmania: This method became a bit too convoluted IMO, it's hard to follow the logic with multiple indentations and icon + tooltip logic mixed together. We could do the icon logic separately from the tooltip logic. I also don't like how "Playing" or "Game Lobby" are duplicated - let's give it some descriptive name like "tooltip_header"and add info like 'live delay' or 'private game' separately, maybe even on a separate line.

Actually, maybe we can operate on lists of lines at first, then just join the list with `</br>`.

src/chat/chatter.py
GrotheFAF: did that

src/chat/chatter.py
GrotheFAF: sorry missed that

src/chat/chatter.py
GrotheFAF: nope, was ok .. but admitted complicated

src/chat/chatter.py
GrotheFAF: I rearanged it with header und game separate and in order

src/chat/chatter.py
```diff
@@ -248,19 +250,16 @@ def updateUser(self):
         self._verifySortOrder()
 
     def updatePlayer(self):
-        self.set_color()
```
Wesmania, line 251: Why was this removed?

src/chat/chatter.py
```diff
@@ -273,43 +272,49 @@ def updatePlayer(self):
...
+                if game.has_live_replay:
+                    icon_str, tooltip_header_str = "playing", "<b>Playing" + private_str + "</b>"
+                else:
+                    icon_str, tooltip_header_str = "playing5", "<b>Playing" + private_str + "</b> - LIVE DELAY (5 Min)"
+            else:  # game.state == something else
+                icon_str, tooltip_header_str = "unknown", "<b>Playing maybe ...</b>"
+
+            formatting = "<br/>title: {}<br/>mod: {}<br/>map: {}<br/>players: {} / {}<br/>id: {}"
+            tooltip_game_str = formatting.format(game.title, game.featured_mod, maps.getDisplayName(game.mapname),
+                                                 str(game.num_players), str(game.max_players), str(game.uid))
```
Wesmania: format() does conversion to string for us.

src/chat/chatter.py
```diff
@@ -273,43 +272,49 @@ def updatePlayer(self):
...
-        self._verifySortOrder()
+            icon_str = "newplayer"
+        self.rankItem.setIcon(util.THEME.icon("chat/rank/%s.png" % icon_str))
+        self.rankItem.setToolTip(tooltip_str)
 
     def updateGame(self):
         # Status icon handling
         game = self.user_game
-        player = self.user_player
         if game is not None and not game.closed():
```
Wesmania: I'd still like to see icon_str moved into a separate method. I know it seems like repeating all these cases, but it will really make things more readable (and in the future we might just want to change the tooltip separately from the icon anyway)

src/chat/chatter.py
```diff
@@ -273,43 +272,49 @@ def updatePlayer(self):
...
             elif game.state == GameState.PLAYING:
-                self.statusItem.setIcon(util.THEME.icon("chat/status/playing.png"))
-                self.statusItem.setToolTip("Playing Game<br/>"+url.toString())
-        else:
-            self.statusItem.setIcon(QtGui.QIcon())
-            self.statusItem.setToolTip("Idle")
+                if game.has_live_replay:
+                    icon_str, tooltip_header_str = "playing", "<b>Playing" + private_str + "</b>"
+                else:
+                    icon_str, tooltip_header_str = "playing5", "<b>Playing" + private_str + "</b> - LIVE DELAY (5 Min)"
```
Wesmania: Python's format() accepts keywords, and you can do stuff like "game {private}".format(private=foo, delay=bar) with no errors from the extra argument. If we set tooltip_header_str, private_str and delay string separately, then just format() the header with both, we might reduce string duplication a bit more.

src/chat/chatter.py
```diff
@@ -432,12 +437,12 @@ def send_the_orcs():
                 menu_add("Join hosted Game", self._interactWithGame, True)
             elif game.state == GameState.PLAYING:
                 time_running = time.time() - game.launched_at
-                if time_running > 5 * 60:
+                if time_running > game.LIVE_REPLAY_DELAY_SECS:
```
Wesmania: We can use game.has_live_replay here.

GrotheFAF: yes we can.

src/chat/chatter.py
GrotheFAF: aren't we lucky ;-)

src/chat/chatter.py
GrotheFAF: packed it in one.

src/chat/chatter.py
GrotheFAF: not needed (see description)

src/chat/chatter.py
```diff
@@ -273,47 +272,73 @@ def updatePlayer(self):
...
+        self.update_status_icon()
+        self.update_map()
+
+    def update_status_tooltip(self):
+        # Status tooltip handling
+        game = self.user_game
+        if game is not None and not game.closed():
+            private_str = " (private)" if game.password_protected else ""
+            if game.state == GameState.OPEN:
+                head_str = "Hosting" if game.host == self.user.name else "In Lobby"
```
Wesmania: The head_str and add_str feel like something better expressed with formatting. We could do something like `header = "Hosting {private}game:"`, `header = In {private}lobby:`, `Playing{delay}:`, then just do `header.format(private=foo, delay=bar)` and use that for the second formatting.

I'm sorry for being so nitpicky, but I still have nightmares from refactoring the game tooltip string functions :( One day we're just going to use jinja2 for all this crap.

GrotheFAF: come on you like it ;-)

src/chat/chatter.py
```diff
@@ -273,47 +272,76 @@ def updatePlayer(self):
...
+
+    def update_status_tooltip(self):
+        # Status tooltip handling
+        game = self.user_game
+        if game is not None and not game.closed():
+            private_str = "(private)" if game.password_protected else ""
+            delay_str = ""
+            if game.state == GameState.OPEN:
+                if game.host == self.user.name:
+                    head_str = "Hosting {private} game</b>"
```
Wesmania: Move one of spaces to private_str so we don't get double space when it's empty.

src/chat/chatter.py
```diff
@@ -273,47 +273,76 @@ def updatePlayer(self):
...
 
         league = player.league
         if league is not None:
-            formatting = ("Division : {}\n"
-                          "Global Rating: {}")
-            self.rankItem.setToolTip(formatting.format(
-                league["division"],
-                str(int(rating))))
-            self.rankItem.setIcon(util.THEME.icon("chat/rank/%s.png" % league["league"]))
+            icon_str, tooltip_str = league["league"], "Division : " + league["division"] + "\n" + tooltip_str
```
Wesmania: Was removing rating from tooltip intentional here? also format() plox :)

src/chat/chatter.py
```diff
@@ -273,47 +273,76 @@ def updatePlayer(self):
...
         else:
-            self.rankItem.setIcon(util.THEME.icon("chat/rank/newplayer.png"))
-
-        self._verifySortOrder()
+            icon_str = "newplayer"
+        self.rankItem.setIcon(util.THEME.icon("chat/rank/%s.png" % icon_str))
```
Wesmania: format(), for consistency.

Wesmania (commented):
Last nitpicks, after that just make a changelog entry and I'll merge.

src/chat/chatter.py
GrotheFAF: guess what's in  + tooltip_str (at the end)
... and I remembered: 'format() does conversion to string for us'

GrotheFAF:
add changelog

PR #893, opened 2017-12-01
add 'View in Leaderboards' to chat user context menu

GrotheFAF:


src/chat/chatter.py
```diff
@@ -482,3 +487,15 @@ def joinInGame(self, url):
 
     def viewReplay(self, url):
         replay(url)
+
+    def view_in_leaderboards(self):
```
Wesmania: I can't accept all this code being put in the chatter class, what you're doing here is responsibility of the tournament tab, not the chatter. Put tournament tab code in the tournament tab and expose an interface to it, refactor code if necessary to do so. Same goes for client and accessing / switching tabs, details of it should be in the client class and exposed to the outside as functions, maybe we can expose tabs as member of some enum.

Once there, make sure we don't use magic numbers or URLs, put them in constants or in settings.

src/chat/chatter.py
```diff
@@ -426,6 +426,11 @@ def send_the_orcs():
         # Aliases link
         menu_add("View Aliases", self.viewAliases, True)
 
+        # Show in Leaderboards
```
Wesmania: I think comments are not needed here, the code documents itself via variable names.

Wesmania (changes requested):

src/chat/chatter.py
GrotheFAF: tournament ?

Wesmania: Leaderboard tab, my mistake.

GrotheFAF: I don't understand what you want,
 could you explain it on how it is done on viewVaultReplay (some lines above)?

Wesmania: The code in viewVaultReplay is also really ugly, but there's no reason to perpetuate this ugliness. Point is (taking viewVaultReplay as example), the chatter should not need to know about private details like replays.setCurrentIndex or replays.vaultManager. This is old code that violates principle of single responsibility.

What we should do is put methods that do leaderboards stuff in the whatever class 'leaderboards' is, and call it with whatever it needs (in this case, the user). Same for the tab setting in client, there should be a method that sets the tab (maybe based on some enum) rather than directly manipulating its mainTabs member (and while we're at it, some nicer way of grabbing tabs too, maybe also based on some enum).

src/chat/chatter.py
GrotheFAF: well, all others have one .. so ..

src/chat/chatter.py
GrotheFAF: well, I tried some ...

GrotheFAF: what about: 'self.chat_widget.client....' visa 'client.instance....'  ?
(sheo once recommented  to use the later throughout)

Wesmania: It doesn't matter, the thing that matters is that we don't fiddle with other objects' (de facto) private members, and instead move any parts of this code that do into these objects as methods.

GrotheFAF: so in practice chatter should call a func in _clientwindow which calls a function in _replaywidget?

Wesmania: It's going in the right direction, I'd add a dedicated method in _clientwindow for setting active tab (maybe using some enum instead of numbers), and a dedicated 'show this user in leaderboards' method in leaderboards widget (cramming it into busy_entered is not a good idea, since busy_entered is meant to be called only when we switch to the tab). Both should be called from the chatter - we can grab the tournament widget directly from client as attribute, or add a method in client for accessing different tabs if it would feel prettier.

GrotheFAF: guess that should do it.

src/stats/_statswidget.py
```diff
@@ -187,25 +188,29 @@ def _injectCSS(self):
         if util.THEME.themeurl("ladder/style.css"):
             injectWebviewCSS(self.webview.page(), util.THEME.readstylesheet("ladder/style.css"))
 
+    def set_player(self, player):
+        if self.player != player:
+            self.player = player
+            self.loaded = False
```
Wesmania: Optional nitpicks - self.player -> self.selected_player and self.loaded -> self.should_reload would document themselves a bit better IMO.

Wesmania (approved):
Looks good. Add changelog entry, fix optional nitpicks if you feel like it, and I'll merge.

GrotheFAF:
added changelog

src/stats/_statswidget.py
```diff
@@ -187,25 +188,29 @@ def _injectCSS(self):
...
-            self.leagues.setCurrentIndex(me.league - 1)
+        if self.selected_player is None:
+            self.selected_player = self.client.players[self.client.login]
+        if self.selected_player.league is not None:
+            self.leagues.setCurrentIndex(self.selected_player.league - 1)
         else:
             self.leagues.setCurrentIndex(5)  # -> 5 = direct to Ladder Ratings
 
-        if self.loaded:
+        if self.selected_loaded:
```
Wesmania: selected_loaded is less clear IMO, loaded or should_reload would be much better.

Wesmania (commented):
typo in commit message

src/stats/_statswidget.py
GrotheFAF: should_reload would be the invers logic, maybe: selected_player_loaded ?

Wesmania: selected_player_changed would be one way to call it. It really depends on how we'd extend it if there were more things to update. I'd do it by having each one set this variable to false, then check only this variable to see if we need to reload (hence should_reload makes sense). Your approach is to have each updated thing set its own variable, then check and reset all of them when reloading, so selected_player_changed fits better. Pick whichever you think is best, both are pretty good.

GrotheFAF: selected_player_changed is again invers logic,
 which had to be set to True in '__init__' which feels wrong to me.

GrotheFAF:
found the typo...

Issue #894, opened 2017-12-02
lose the 'as..' part of Import Xwidget as X


GrotheFAF:
I think it's clearer, less confusing and consistend to refer to widgets as widgets.

example:
from stats.__init__:
from ._statswidget import StatsWidget as Stats
in _clientwindow:
self.ladder = stats.Stats(self)

would be clearer:
from ._statswidget import StatsWidget
self.ladder = stats.StatsWidget(self)

Discuss ...

Wesmania:
+1

GrotheFAF:
There are 2 cases of 'instance' that are used all over:
client\__init__: instance = ClientWindow()
fa\game_process: instance = GameProcess()

might be usefull to rename the later to: game_instance = GameProcess()

PR #897, opened 2017-12-04
remove import Xwidget as X - 'fix' #894

GrotheFAF:
mostly in __init__.py used in _clientwindow.py

### DIFF
--- a/src/fa/game_process.py
+++ b/src/fa/game_process.py
@@ -150,4 +150,5 @@ class GameProcess(QtCore.QProcess):
 
             self.close()
 
+
 instance = GameProcess()
--- a/src/fa/game_session.py
+++ b/src/fa/game_session.py
@@ -5,7 +5,7 @@ from enum import IntEnum
 from connectivity.turn import TURNState
 from decorators import with_logger
 from fa.game_connection import GPGNetConnection
-from fa.game_process import GameProcess, instance as game_process_instance
+from fa.game_process import instance
 
 
 class GameSessionState(IntEnum):
@@ -64,7 +64,7 @@ class GameSession(QObject):
         # We only allow one game connection at a time
         self._game_connection = None
 
-        self._process = game_process_instance  # type: GameProcess
+        self._process = instance  # type:'GameProcess'
         self._process.started.connect(self._launched)
         self._process.finished.connect(self._exited)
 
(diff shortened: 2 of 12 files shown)


src/fa/game_process.py
```diff
@@ -97,7 +97,7 @@ def run(self, info, arguments, detach=False, init_file=None):
             logger.info("Running FA via executable: " + executable)
 
             # Launch the game as a stand alone process
-            if not instance.running():
```
Wesmania: I'm surprised that this code used to run, I can't find any 'instance' import earlier in the file. There must be some ugly importing done earlier somewhere. We should explicitly import fa so it doesn't mysteriously break in the future.

src/fa/game_session.py
```diff
@@ -5,7 +5,7 @@
 from connectivity.turn import TURNState
 from decorators import with_logger
 from fa.game_connection import GPGNetConnection
-from fa.game_process import GameProcess, instance as game_process_instance
```
Wesmania, line 8: My linter complains that GameProcess is unused, might fix that too while we're at it.

Wesmania (approved):
Looks good to me, just minor optional stuff.

src/fa/game_process.py
GrotheFAF: oi .. I might have been sleeping doing that  .. I can't remember putting that 'fa.' there, because 'fa.instance.running' doesn't reference here. 
And no mystery because 'instance' is defined here line 154 ;-)

src/fa/game_session.py
GrotheFAF: and your linter is right, pycharm was fooled by the type comment in line 67

GrotheFAF: and while we're at it .. what does your linter say about:
L9 'import chat'  and  L56 'import sys' in _clientwindow.py?
we welcome less importing, right?

Wesmania: It complains about these and a couple more lines. I'd rather avoid mass pruning these just yet to make rebasing the ice branch easier, this one is just in the way anyway. It'd be trivial to do either way, I'm using flake8 as a linter and I'm sure there are ways to automatically prune the stuff.

GrotheFAF: still complains here, after last update?

Wesmania: It's all good now.

GrotheFAF: pycharm is unhappy with the 3  @_needs_game_connection
it would be happy with @_needs_game_connection()
is that a thing, or just stupid pycharm?

PR #898, opened 2017-12-05
don't redo myTree if nothing has changed

GrotheFAF:
As updatemyTree is called on every change to the main Replay tab and if there are a lot of local replays, this will avoid useless wait time. (slower computer exist)

remember modification_time of REPLAY_DIR to avoid redoing myTree if nothing has changed

+ minor cleanup in ReplayWidget

### DIFF
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -233,6 +233,7 @@ class LocalReplaysWidgetHandler(object):
         self.myTree.header().setSectionResizeMode(1, QtWidgets.QHeaderView.ResizeToContents)
         self.myTree.header().setSectionResizeMode(2, QtWidgets.QHeaderView.Stretch)
         self.myTree.header().setSectionResizeMode(3, QtWidgets.QHeaderView.ResizeToContents)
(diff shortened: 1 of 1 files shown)


src/replays/_replayswidget.py
```diff
@@ -273,6 +274,11 @@ def myTreeDoubleClicked(self, item):
             replay(item.filename)
 
     def updatemyTree(self):
+        modification_time = os.path.getmtime(util.REPLAY_DIR)
+        if self.myTree.modification_time < modification_time:  # anything changed?
```
Wesmania: Just a style thing, I'd reverse branch order (>= instead) - check if we should bail out first. Lets us put the modification_time setting line outside of condition branches, too.

Wesmania (approved):
Looks good, just stylistic nitpicks if you feel like it. I'll merge tomorrow morning regardless.

src/replays/_replayswidget.py
GrotheFAF: yes, better. (but is == then)

PR #900, opened 2017-12-05
fix #783 live game ended online-replay-doubleclick

### LINKED ISSUE #783: finished live replays cant be loaded

GrotheFAF:
This is because the server also sends live games with a replay search. (bug or feature?)

if a search gets old, shown 'playing' games might already have ended,
which leads to a bad click experience, we don't want that.

check if live-game is still running, else give option to start replay, fix #783
check for player still in game and playerset to join live game

### DIFF
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -421,10 +421,11 @@ class ReplayVaultWidgetHandler(object):
     automatic = Settings.persisted_property("replay/automatic", default_value=False, type=bool)
     spoiler_free = Settings.persisted_property("replay/spoilerFree", default_value=True, type=bool)
 
-    def __init__(self, widget, dispatcher, client, playerset):
+    def __init__(self, widget, dispatcher, client, gameset, playerset):
         self._w = widget
         self._dispatcher = dispatcher
         self.client = client
+        self._gameset = gameset
         self._playerset = playerset
 
         self.onlineReplays = {}
@@ -501,19 +502,31 @@ class ReplayVaultWidgetHandler(object):
                     item.generateInfoPlayersHtml()
 
     def onlineTreeDoubleClicked(self, item):
-        if hasattr(item, "duration"):
+        if hasattr(item, "duration"):  # it's a game not a date separator
             if "playing" in item.duration:  # live game will not be in vault
-                if not item.live_delay:  # live game under 5min
-                    name = None
-                    if item.mod == "ladder1v1":
-                        name = item.name[:item.name.find(" ")]  # "name vs name"
-                    else:
-                        players = (p for t in item.teams for p in t)
-                        for player in players:
-                            name = player["name"]
-                            if name != "" and name in self._playerset:
+                # search result isn't updated automatically - so game status might have changed
+                if item.uid in self._gameset.games:  # game still running
+                    game = self._gameset.games[item.uid]
+                    if not game.launched_at:  # we frown upon those
+                        return
+                    if game.has_live_replay:  # live game over 5min
+                        for name in game.players:  # find a player ...
+                            if name in self._playerset:  # still logged in
+                                self._startReplay(name)
                                 break
-                    self._startReplay(name)
+                    else:
+                        wait_str = time.strftime('%M Min %S Sec', time.gmtime(game.LIVE_REPLAY_DELAY_SECS -
+                                                                              (time.time() - game.launched_at)))
+                        QtWidgets.QMessageBox.information(client.instance, "5 Minute Live Game Delay",
+                                                          "It is too early to join the Game.\n"
+                                                          "You have to wait " + wait_str + " to join.")
+                else:  # game ended - ask to start replay
+                    if QtWidgets.QMessageBox.question(client.instance, "Live Game ended",
+                                                      "Would you like to watch the replay from the vault?",
+                                                      QtWidgets.QMessageBox.Yes,
+                                                      QtWidgets.QMessageBox.No) == QtWidgets.QMessageBox.Yes:
+                        self.replayDownload.get(QNetworkRequest(QtCore.QUrl(item.url)))
+
             else:  # start replay
                 if hasattr(item, "url"):
                     self.replayDownload.get(QNetworkRequest(QtCore.QUrl(item.url)))
@@ -630,7 +643,7 @@ class ReplaysWidget(BaseClass, FormClass):
 
         self.liveManager = LiveReplaysWidgetHandler(self.liveTree, client, gameset)
         self.localManager = LocalReplaysWidgetHandler(self.myTree)
-        self.vaultManager = ReplayVaultWidgetHandler(self, dispatcher, client, playerset)
+        self.vaultManager = ReplayVaultWidgetHandler(self, dispatcher, client, gameset, playerset)
 
         logger.info("Replays Widget instantiated.")
 


src/replays/_replayswidget.py
```diff
@@ -501,19 +501,24 @@ def onlineTreeClicked(self, item):
...
-                    name = None
-                    if item.mod == "ladder1v1":
-                        name = item.name[:item.name.find(" ")]  # "name vs name"
-                    else:
-                        players = (p for t in item.teams for p in t)
-                        for player in players:
-                            name = player["name"]
-                            if name != "" and name in self._playerset:
+                # search result isn't updated automatically - so game status might have changed
+                if item.uid in client.instance.gameset.games:  # game still running
```
Wesmania: I'd rather we passed gameset through the constructor the same way we do with playerset.

src/replays/_replayswidget.py
```diff
@@ -501,19 +501,24 @@ def onlineTreeClicked(self, item):
...
+                    game = client.instance.gameset.games[item.uid]
+                    if not game.launched_at:  # we frown upon those
```
Wesmania: 'game.state == OPEN' seems to check what we want to check more directly.

src/replays/_replayswidget.py
```diff
@@ -501,19 +501,24 @@ def onlineTreeClicked(self, item):
...
+                        return
+                    if game.has_live_replay:  # live game over 5min
+                        for name in game.players:  # find a player ...
+                            if name in self._playerset:  # still logged in
+                                self._startReplay(name)
                                 break
-                    self._startReplay(name)
+                else:  # game ended - ask to start replay
+                    if QtWidgets.QMessageBox.question(client.instance, "Live Game ended", "Want to try the Replay?",
```
Wesmania: People usually look for games that already ended when they use the replay vault, I think there's no need to ask them anything.

src/replays/_replayswidget.py
```diff
@@ -501,19 +501,24 @@ def onlineTreeClicked(self, item):
...
-                        players = (p for t in item.teams for p in t)
-                        for player in players:
-                            name = player["name"]
-                            if name != "" and name in self._playerset:
+                # search result isn't updated automatically - so game status might have changed
+                if item.uid in client.instance.gameset.games:  # game still running
+                    game = client.instance.gameset.games[item.uid]
+                    if not game.launched_at:  # we frown upon those
+                        return
+                    if game.has_live_replay:  # live game over 5min
```
Wesmania: I often find myself double-clicking the 'crossed swords' icon in chat before the 5 minute period ends and just waiting for the game to start playing the replay. I think it might be a use case here too. I don't really know what would be the best way to handle it, maybe some popup? What do you think?

src/replays/_replayswidget.py
GrotheFAF: you mean add gameset to ReplayVaultWidgetHandler ?

src/replays/_replayswidget.py
GrotheFAF: 'we' don't want that. (no open game would show up in replays). It's really just about missing launch-time. rare ... but happened.

src/replays/_replayswidget.py
GrotheFAF: yeah though about that .. a: 'too early .. wait n:nn to watch the live game'
(but I always hesitate to complicate things .. cause then you brakedown and cry ;-)

src/replays/_replayswidget.py
GrotheFAF: yes, but this is the case were the list is still showing 'playing'

GrotheFAF:
the whole thing is about handling that the server sends also live games on a replay request. (lets call it a bug-around). Once that is fixed ..this part can be removed ... so this should be temporary.

src/replays/_replayswidget.py
Wesmania: Alright. Please have the comment mention that, then.

src/replays/_replayswidget.py
Wesmania: Well, an alternative to a popup is an angry user not knowing why the game doesn't launch even though he double-clicked. Alternatively we can just proceed with launch anyway, people might be already familiar with how launching delayed games works.

src/replays/_replayswidget.py
Wesmania: Alright, good point.

src/replays/_replayswidget.py
Wesmania: Yeah, the same way playerset is added. We don't even have to look far for it, since ReplaysWidget has it.

Wesmania:
Yeah, I see your point, we need to handle occasional discrepancies between what server tells us here and what we know about current games. It'd be good to document that in the comments.

src/replays/_replayswidget.py
GrotheFAF: I thought  'we frown upon those' said enough ;-)

src/replays/_replayswidget.py
GrotheFAF: ok, done.
(but you recently wrote - when I asked about 'self.....X' visa 'client.instant...X' - that it doesn't matter)

src/replays/_replayswidget.py
GrotheFAF: I'll make a msg.

src/replays/_replayswidget.py
Wesmania: Not in 3 months when we revisit this code and wonder what the flying heck 'frowning at games' means :/

src/replays/_replayswidget.py
Wesmania: Client is a shoggoth class that we can't hope to make manageable, so it doesn't matter that much. These however are shiny new model classes, so proper dependency injection is very much in fashion :)

GrotheFAF:
ok I polished that turd .. shiny again ;-)

src/replays/_replayswidget.py
```diff
@@ -501,19 +502,30 @@ def onlineTreeClicked(self, item):
...
+                        for name in game.players:  # find a player ...
+                            if name in self._playerset:  # still logged in
+                                self._startReplay(name)
                                 break
-                    self._startReplay(name)
+                    else:
+                        wait_str = time.strftime('%M Min %S Sec', time.gmtime(game.LIVE_REPLAY_DELAY_SECS -
+                                                                              (time.time() - game.launched_at)))
+                        QtWidgets.QMessageBox.information(client.instance, "5 Minute Live Game Delay",
+                                                          "This Game is still too young to join.\n"
```
Wesmania: "too young" sounds a bit weird. Maybe "The game started recently and it's too early to spectate"?

src/replays/_replayswidget.py
```diff
@@ -501,19 +502,30 @@ def onlineTreeClicked(self, item):
...
+                                                          "You have to wait " + wait_str + " to join.")
+                else:  # game ended - ask to start replay
+                    if QtWidgets.QMessageBox.question(client.instance, "Live Game ended", "Want to try the Replay?",
```
Wesmania: I'd go with "Would you like to watch a replay from the vault?"

src/replays/_replayswidget.py
GrotheFAF: I have a poetic licence ;-)
"spectate" sounds to sexual ... and that the game "started recently" is obvious, it says so in the list.
would "too early" be ok?

Wesmania: I guess so, something like "It's too early yet to watch this game".

src/replays/_replayswidget.py
GrotheFAF: well, the user double clicked on a specific game - so the message refered to that game - hence "... the Replay" of that game the user clicked. (and "try" because replays can be corrupted, hence not starting)

Wesmania: I know, I just think a bit more formal style goes better with all other strings in the client, "Want to try a replay?" is a little bit too direct IMO.

GrotheFAF: Might be, people here are so easy offended ;-)

Wesmania: It's not that, it's a matter of having an elegant UI. I deal with enough broken korglish daily, I want our client to look 100% professional in this regard :)

Wesmania:
Everything's good. It's pretty hard for me to test that everything works, though, can't find any games in the vault that are in a "playing" state, so I'll wait with merging just a little bit. Did you verify that it works?

GrotheFAF:
yes ... to test it, just take a player from chat with red-crossed-swords and context-menu -> View replays in vaut .. there you go... (most search IN vault gives you live games)

GrotheFAF:
would be nice to have the live games with a live duration clock .. tick tick ...

Wesmania:
Alright, looks good, merging. I'd give some strings a do-over, but it's minor and nitpickable in the future.

Issue #899, opened 2017-12-05
#newbie chat channel autojoin


JaggedAppliance:
http://forums.faforever.com/viewtopic.php?f=2&t=15555

I have registered a channel for newbies. Would it be possible to have players with less than (for example) 50 games played autojoined to this channel?

Viba:
_chatwidget.py line 64, adding #newbie there makes everyone using python client autojoin.
Include filter for amount of games with "self.number_of_games" if #newbie should be added or not?

JaggedAppliance:
https://github.com/FAForever/client/pull/357

Perhaps this could be salvaged?

ghost:
Yes salvaged it. Don't know what the criteria should be used to distinguish newbie or not.. so I used deviation (like the game lobby with the white rating) : deviation > 100 (on ladder or global) => considered newbie

EDIT : Nevermind, using deviation is highly related to the number of games but someone who only play ladder or global will always be considered newbie.. so switching to number of games, I guess 50?

PR #901, opened 2017-12-05
Add autojoin #newbie feature and auto join channels

ghost:
I just copy & pasta the old PR from https://github.com/FAForever/client/issues/899 and edited a bit.
It seems to work but I would like to see on someone else computer which is considered a newbie (number of games < 51) if they autojoin by default the newbie channel or if they need to activate in the option first (which shouldn't be the case and fixed if it is).

In the option you can check or uncheck autojoin newbie channel. There is also another feature where you can add channels you want to autojoin (for example if you are #poker , you can then go to the setting and add this channel to autojoin). Feel free or suggestion or anything. 

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/src/chat/_chatwidget.py
+++ b/src/chat/_chatwidget.py
@@ -32,6 +32,8 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
     irc_host = Settings.persisted_property('chat/host', type=str, default_value='irc.' + defaults['host'])
     irc_tls = Settings.persisted_property('chat/tls', type=bool, default_value=False)
 
+    auto_join_channels = Settings.persisted_property('chat/auto_join_channels', type=list, default_value=[])
+
     """
     This is the chat lobby module for the FAF client.
     It manages a list of channels and dispatches IRC events (lobby inherits from irclib's client class)
@@ -257,6 +259,11 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
         # Perform any pending autojoins (client may have emitted autoJoin signals before we talked to the IRC server)
         self.autoJoin(self.optionalChannels)
         self.autoJoin(self.crucialChannels)
+        self.autoJoin(self.auto_join_channels)
+
+        max_number_of_games_to_be_considered_newbie = 51
+        if self.client.useNewbiesChannel and self.client.me.player.number_of_games < max_number_of_games_to_be_considered_newbie:
+            self.autoJoin(["#newbie"])
 
     def nickservRegister(self):
         if hasattr(self, '_nickserv_registered'):
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -41,6 +41,7 @@ Created on Dec 1, 2011
 '''
 
 from PyQt5 import QtCore, QtWidgets, QtGui
+from PyQt5.QtWidgets import QInputDialog, QLineEdit
 
 from client import ClientState, LOBBY_HOST, \
     LOBBY_PORT, LOCAL_REPLAY_PORT
@@ -767,6 +768,8 @@ class ClientWindow(FormClass, BaseClass):
         self.actionSetSoundEffects.triggered.connect(self.updateOptions)
         self.actionSetOpenGames.triggered.connect(self.updateOptions)
         self.actionSetJoinsParts.triggered.connect(self.updateOptions)
+        self.actionSetNewbiesChannel.triggered.connect(self.updateOptions)
+        self.actionSetAutoJoinChannels.triggered.connect(self.show_autojoin_settings_dialog)
         self.actionSetLiveReplays.triggered.connect(self.updateOptions)
         self.actionSetChatMaps.triggered.connect(self.toggleChatMaps)
         self.actionSaveGamelogs.toggled.connect(self.on_actionSavegamelogs_toggled)
@@ -784,6 +787,7 @@ class ClientWindow(FormClass, BaseClass):
         self.soundeffects = self.actionSetSoundEffects.isChecked()
         self.game_announcer.announce_games = self.actionSetOpenGames.isChecked()
         self.joinsparts = self.actionSetJoinsParts.isChecked()
+        self.useNewbiesChannel = self.actionSetNewbiesChannel.isChecked()
         self.chatmaps = self.actionSetChatMaps.isChecked()
         self.game_announcer.announce_replays = self.actionSetLiveReplays.isChecked()
 
@@ -870,12 +874,26 @@ class ClientWindow(FormClass, BaseClass):
         util.settings.setValue("geometry", self.saveGeometry())
         util.settings.endGroup()
 
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = []
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
+        text_of_autojoin_settings_dialog = """
+        Enter the list of channels you want to autojoin at startup, separated by ;
+        For example: #poker;#newbie
+        To disable autojoining channels, leave the box empty and press OK.
+        """
+        channels_input_of_user, ok = QInputDialog.getText(self, 'Set autojoin channels',
+            text_of_autojoin_settings_dialog, QLineEdit.Normal, ';'.join(autojoin_channels_list))
+        if ok:
+            Settings.set('chat/auto_join_channels', list(map(str.strip, channels_input_of_user.split(';'))))
+
     def saveChat(self):
         util.settings.beginGroup("chat")
         util.settings.setValue("soundeffects", self.soundeffects)
         util.settings.setValue("livereplays", self.game_announcer.announce_replays)
         util.settings.setValue("opengames", self.game_announcer.announce_games)
         util.settings.setValue("joinsparts", self.joinsparts)
+        util.settings.setValue("newbiesChannel", self.useNewbiesChannel)
         util.settings.setValue("chatmaps", self.chatmaps)
         util.settings.setValue("coloredNicknames", self.player_colors.coloredNicknames)
         util.settings.setValue("friendsontop", self.friendsontop)
@@ -903,6 +921,7 @@ class ClientWindow(FormClass, BaseClass):
             self.game_announcer.announce_replays = (util.settings.value("livereplays", "true") == "true")
             self.player_colors.coloredNicknames = (util.settings.value("coloredNicknames", "false") == "true")
             self.friendsontop = (util.settings.value("friendsontop", "false") == "true")
+            self.useNewbiesChannel = (util.settings.value("newbiesChannel","true") == "true")
 
             util.settings.endGroup()
             self.actionColoredNicknames.setChecked(self.player_colors.coloredNicknames)
@@ -912,6 +931,7 @@ class ClientWindow(FormClass, BaseClass):
             self.actionSetOpenGames.setChecked(self.game_announcer.announce_games)
             self.actionSetJoinsParts.setChecked(self.joinsparts)
             self.actionSetChatMaps.setChecked(self.chatmaps)
+            self.actionSetNewbiesChannel.setChecked(self.useNewbiesChannel)
         except:
             pass
 
--- a/changelog.md
+++ b/changelog.md
@@ -18,6 +18,7 @@
  * Add host & live-game-delay to user status + nicer tooltip in Chat (#892)
  * Add a button to see the ladder maps pool from play tab (#882)
  * Add 'view in leaderboard' to chat user context menu (#893)
+ * Add autojoin #newbie channel and the possibility to autojoin other channels (#899)
 
 Contributors:
  - Wesmania
--- a/res/client/client.ui
+++ b/res/client/client.ui
@@ -282,6 +282,8 @@
      </property>
      <addaction name="actionColoredNicknames"/>
      <addaction name="actionFriendsOnTop"/>
+     <addaction name="actionSetNewbiesChannel"/>
+     <addaction name="actionSetAutoJoinChannels"/>
      <addaction name="actionSetJoinsParts"/>
      <addaction name="actionSetOpenGames"/>
      <addaction name="actionSetLiveReplays"/>
@@ -400,6 +402,19 @@
    </property>
    <property name="text">
     <string>&amp;Automatic Login</string>
+   </property>
+  </action>
+   <action name="actionSetNewbiesChannel">
+   <property name="checkable">
+    <bool>true</bool>
+   </property>
+   <property name="text">
+    <string>Use Newbies Channel</string>
+   </property>
+  </action>
+   <action name="actionSetAutoJoinChannels">
+   <property name="text">
+    <string>Set autojoin channels</string>
    </property>
   </action>
   <action name="actionSetAutoDownloadMods">


src/chat/_chatwidget.py
```diff
@@ -32,6 +32,8 @@ class ChatWidget(FormClass, BaseClass, SimpleIRCClient):
     irc_host = Settings.persisted_property('chat/host', type=str, default_value='irc.' + defaults['host'])
     irc_tls = Settings.persisted_property('chat/tls', type=bool, default_value=False)
 
+    autoJoinChannels = Settings.persisted_property('chat/autoJoinChannels', type=list, default_value=[])
```
Wesmania: I'd prefer we used under_scores rather than camelCase.

src/chat/_chatwidget.py
```diff
@@ -231,6 +233,35 @@ def join(self, channel):
     def log_event(self, e):
         self.serverLogArea.appendPlainText("[%s: %s->%s]" % (e.eventtype(), e.source(), e.target()) + "\n".join(e.arguments()))
 
+    def updateAutoJoinIRCMenu(self):
+        self.client.menuAutoJoinIRC.clear()
+
+        # add channels we are currently in
+        for channel in self.channels:
```
Wesmania: I think it would be more reasonable to either make this list static rather than generated dynamically, or let the user specify a list of names directly in some way (e.g. comma-separated).

src/chat/_chatwidget.py
ghost: Ok

src/chat/_chatwidget.py
ghost: As you prefer, to be honest I liked this dynamic one (because very simple to use except a few click) but that's fine, will see how I can edit it to make it not dynamic like this. Probably something in the menu where when you click, it shows a popup or something asking to enter the channels you want to autojoin, and then you will see the list of these channels, but then need the possibility to remove them too.. mmh.

Wesmania: Well, we can always just pre-populate the popup with existing autojoin channels. I'm not too fond of dynamic menu, especially dependent on current channels - it makes configuration confusingly dependent on what channels we're in, and prevents us from changing settings of channels we left unless we rejoin them.

ghost:
What about now @Wesmania ?

src/chat/_chatwidget.py
ghost: I see, I did another solution instead of this dynamic one. Basically you click on set autojoin channels then popup, and you can write what you want to join, and you can also see the list.

src/chat/_chatwidget.py
```diff
@@ -257,6 +259,10 @@ def on_identified(self):
         # Perform any pending autojoins (client may have emitted autoJoin signals before we talked to the IRC server)
         self.autoJoin(self.optionalChannels)
         self.autoJoin(self.crucialChannels)
+        self.autoJoin(self.auto_join_channels)
+
+        if self.client.useNewbiesChannel and self.client.me.player.number_of_games < 51:
```
Wesmania: Give a name to the '51' constant. Dunno where best to put it, you can put it statically in this class for now I guess, or in global scope.

src/client/_clientwindow.py
```diff
@@ -41,6 +41,7 @@
 '''
 
 from PyQt5 import QtCore, QtWidgets, QtGui
+from PyQt5.QtWidgets import QInputDialog,QLineEdit
```
Wesmania: Space after comma.

src/client/_clientwindow.py
```diff
@@ -767,6 +768,8 @@ def initMenus(self):
         self.actionSetSoundEffects.triggered.connect(self.updateOptions)
         self.actionSetOpenGames.triggered.connect(self.updateOptions)
         self.actionSetJoinsParts.triggered.connect(self.updateOptions)
+        self.actionSetNewbiesChannel.triggered.connect(self.updateOptions)
+        self.actionSetAutoJoinChannels.triggered.connect(lambda: self.on_actionSetAutoJoinChannels()) #zawa
```
Wesmania: No need to use lambda here, just drop the parentheses. Connecting lambdas to signals is in general a bad idea, since you can never disconnect them.

src/client/_clientwindow.py
```diff
@@ -870,12 +874,18 @@ def saveWindow(self):
         util.settings.setValue("geometry", self.saveGeometry())
         util.settings.endGroup()
 
+    def on_actionSetAutoJoinChannels(self):
```
Wesmania: It'd be nice to give this a more descriptive name. How about 'show_autojoin_settings_dialog'?

src/client/_clientwindow.py
```diff
@@ -870,12 +874,18 @@ def saveWindow(self):
...
+        input, ok = QInputDialog.getText(None, 'Set autojoin channels','Previous list : '+str(Settings.get('chat/auto_join_channels'))+'\nEnter the list of channels you want to join, separated by ;\n\n For example: #poker;#newbie\nTo disable autojoin, leave box empty and press ok', QLineEdit.Normal, "")
```
Wesmania: Setting message box text in a variable first would be a bit more readable - especially since we could use multiple lines for that (after each '\n'). Instead of showing the previous list in the text we can just set dialog box contents.

Also 'input' is a python global function, please rename it.

Wesmania (commented):
Overall looks good, some small things left. I'll test it later.

src/chat/_chatwidget.py
ghost: done

src/client/_clientwindow.py
ghost: done

src/client/_clientwindow.py
ghost: done.

src/client/_clientwindow.py
ghost: Ok, done

src/client/_clientwindow.py
ghost: Done, maybe not sure how you wanted to do it.

ghost:
Did all the changes you asked, feel free if you have any other suggestion

src/client/_clientwindow.py
```diff
@@ -870,12 +874,25 @@ def saveWindow(self):
         util.settings.setValue("geometry", self.saveGeometry())
         util.settings.endGroup()
 
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
```
Wesmania: Add '[]' as a default here, since otherwise we get None.

src/client/_clientwindow.py
```diff
@@ -870,12 +874,25 @@ def saveWindow(self):
...
+        text_of_autojoin_settings_dialog = """
+        Autojoin channels list : """ + str(autojoin_channels_list) + """
```
Wesmania: There's still no need to put the current list separately here. We can pass the default textbox contents through the 'text' parameter, as shown [here](http://doc.qt.io/qt-5/qinputdialog.html#getText). Some ';'.join(list) should be enough.

src/client/_clientwindow.py
```diff
@@ -870,12 +874,25 @@ def saveWindow(self):
...
+        Enter the list of channels you want to join, separated by ;
+        For example: #poker;#newbie
+        To disable autojoin, leave box empty and press ok
```
Wesmania: This gets shown to the user - make sure grammar, interpunction etc. is flawless. IMO we should put dots at sentence end in textboxes.

src/client/_clientwindow.py
```diff
@@ -870,12 +874,25 @@ def saveWindow(self):
...
+        """
+        channels_input_of_user, ok = QInputDialog.getText(None, 'Set autojoin channels',text_of_autojoin_settings_dialog, QLineEdit.Normal, "")
+        if ok:
+            Settings.set('chat/auto_join_channels', channels_input_of_user.split(';'))
```
Wesmania: I think it would be a good idea to strip() the string list so nobody gets confused when they do '#channel1 ; #channel2'. If we really want to keep it a one-liner, we can even do a list(map(str.strip, channels.split(';'))).

Wesmania (commented):
More nitpicks out of testing.

src/client/_clientwindow.py
ghost: ok, done, it is [] as default now

src/client/_clientwindow.py
ghost: Ugh, I think I did what you asked but not sure what you meant...?

src/client/_clientwindow.py
ghost: Ok improved it but I'm not an english native speaker so...Maybe I missed something.
I didn't want to add a point after ; though because people may get confused otherwise and add a . after ; even if there is an example... And same for the example.

src/client/_clientwindow.py
ghost: ok done

ghost:
Did all the changes you asked (have doubt for one two), feel free as usual

src/client/_clientwindow.py
Wesmania: I think it's better to set the textbox contents to current settings rather than describe the settings in the box label. Lose the 'Autojoin channels list:' line, and instead add the joined channel list as the fifth argument to getText (or equivalently, a 'text' keyword argument).

src/client/_clientwindow.py
```diff
@@ -870,12 +874,26 @@ def saveWindow(self):
...
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = []
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
+        text_of_autojoin_settings_dialog = """
+        Autojoin channels list : """ + ';'.join(autojoin_channels_list) + """
+        Enter the list of channels you want to join, separated by ;
+        For example: #poker;#newbie
+        To disable autojoin channels feature, leave the box empty and press ok.
+        """
+        channels_input_of_user, ok = QInputDialog.getText(None, 'Set autojoin channels',text_of_autojoin_settings_dialog, QLineEdit.Normal, "")
```
Wesmania: If you pass the current widget (self) instead of None as the first argument, the dialog will be themed consistently with the rest of the client.

src/client/_clientwindow.py
```diff
@@ -870,12 +874,26 @@ def saveWindow(self):
...
         util.settings.endGroup()
 
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = []
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
+        text_of_autojoin_settings_dialog = """
+        Autojoin channels list : """ + ';'.join(autojoin_channels_list) + """
+        Enter the list of channels you want to join, separated by ;
+        For example: #poker;#newbie
+        To disable autojoin channels feature, leave the box empty and press ok.
```
Wesmania: I'd do "To disable autojoining channels, leave the box empty and press OK."

src/client/_clientwindow.py
```diff
@@ -870,12 +874,26 @@ def saveWindow(self):
         util.settings.setValue("geometry", self.saveGeometry())
         util.settings.endGroup()
 
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = []
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
+        text_of_autojoin_settings_dialog = """
+        Autojoin channels list : """ + ';'.join(autojoin_channels_list) + """
+        Enter the list of channels you want to join, separated by ;
```
Wesmania: I'd put 'autojoin at startup' for 'join' for 120% clarity.

src/client/_clientwindow.py
ghost: Ughh..wut? Sorry I don't get what you mean....

For the fifth argument of getText, I did this : 
channels_input_of_user, ok = QInputDialog.getText(self, 'Set autojoin channels',text_of_autojoin_settings_dialog, QLineEdit.Normal, 'Autojoin channels list : ' + ';'.join(autojoin_channels_list))
But then the line where the user can write the channels has inside this text 'autojoin channels list' which is not good. (added the join channel list as fifth argument to this gettext).

For the box label : 
For me the box label is what is shown to the user, which is the 3rd argument of gettext. And I set the content of the textbox (this popup) to the text inside this variable text_of_autojoin_settings_dialog.
 And what do you mean current settings? what is that?

Wesmania: That first effect you're describing (putting autojoin_channels_list in the box where the user writes stuff) is exactly what I'm aiming for. It feels natural to me that the contents of these box are at the same time the value of the autojoin_channels_list setting. You get this association that what is in the box is the current setting, and what you change is what the new setting will be. It also makes it easier for the user to edit an existing setting - he doesn't have to clumsily copy over the text from the box label to the edited line.

ghost: Ooh, ok ! Got it.

src/client/_clientwindow.py
ghost: done

src/client/_clientwindow.py
ghost: done

src/client/_clientwindow.py
MathieuB8: done

ghost: done, didn't know that it was for that. Was always a mystery this thing.

ghost:
Ok did all the changes you asked, Feel free if you have any other suggestion/idea/something

Wesmania:
Looks good. Squash the commits and wrap that one very long line in _clientwindow.py (to keep them under ~120) and we're good.

ghost:
And done I think @Wesmania

src/client/_clientwindow.py
```diff
@@ -870,12 +874,26 @@ def saveWindow(self):
...
+    def show_autojoin_settings_dialog(self):
+        autojoin_channels_list = []
+        autojoin_channels_list = Settings.get('chat/auto_join_channels')
+        text_of_autojoin_settings_dialog = """
+        Enter the list of channels you want to autojoin at startup, separated by ;
+        For example: #poker;#newbie
+        To disable autojoining channels, leave the box empty and press OK.
+        """
+        channels_input_of_user, ok = QInputDialog.getText(self, 'Set autojoin channels',
+        text_of_autojoin_settings_dialog, QLineEdit.Normal, ';'.join(autojoin_channels_list))
```
Wesmania: Give this line one more indentation level and we're good.

ghost: ok done

ghost:
Ok did all the changes you asked, Feel free if you have any other suggestion/idea/something

Exotic-Retard:
i just want to say this is great stuff. autojoin was something needed for a very long time indeed.
a couple of suggestions:
1. have the newbie channel part of the generic autojoin thing and not by rating, just make it on by default. that way all new players will join it until they turn it off. no need to force anyone to do anything, and they can leave when they feel theyre ready. some people want training even after 100 games for example.
2. this ones much harder - give the ability for people like moderators to set/remove autojoins for other users - would be insanely useful for tournaments where you need to get 25 people to join a tourney channel and not have them ask you for the link every time. if this is hard, then we can just leave it for later, but it would be nice to have.

ghost:
for 1. added an answer in the forum.

for 2 : I was thinking of implementing this kind of feature with the new tourney tab (where you can get info about tourneys) so...

PR #906, opened 2017-12-07
Fixing pinging : you can't ping yourself anymore

ghost:
When you write a message and put your name in it, you won't ping yourself anymore it won't the message with red colour.

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/src/chat/channel.py
+++ b/src/chat/channel.py
@@ -249,9 +249,13 @@ class Channel(FormClass, BaseClass):
         if player.clan is not None:
             displayName = "<b>[%s]</b>%s" % (player.clan, chname)
 
+        sender_is_not_me = chatter.name != self._me.login
+
         # Play a ping sound and flash the title under certain circumstances
         mentioned = text.find(self.chat_widget.client.login) != -1
-        if mentioned or (self.private and not (formatter is Formatters.FORMATTER_RAW and text == "quit.")):
+        is_quit_msg = formatter is Formatters.FORMATTER_RAW and text == "quit."
+        private_msg = self.private and not is_quit_msg
+        if (mentioned or private_msg) and sender_is_not_me:
             self.pingWindow()
 
         avatar = None
@@ -268,7 +272,7 @@ class Channel(FormClass, BaseClass):
             # Fallback and ask the client. We have no Idea who this is.
             color = self.chat_widget.client.player_colors.getUserColor(player.id)
 
-        if mentioned:
+        if mentioned and sender_is_not_me:
             color = self.chat_widget.client.player_colors.getColor("you")
 
         # scroll if close to the last line of the log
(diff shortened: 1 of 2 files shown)


src/chat/channel.py
```diff
@@ -241,9 +241,14 @@ def printLine(self, chname, text, scroll_forced=False, formatter=Formatters.FORM
         if player.clan is not None:
             displayName = "<b>[%s]</b>%s" % (player.clan, chname)
 
+        name_of_author_message=''
+        if hasattr(player, 'login'):
```
Wesmania: There's no need to do it like this if player is a Player instance.

src/chat/channel.py
```diff
@@ -259,7 +264,7 @@ def printLine(self, chname, text, scroll_forced=False, formatter=Formatters.FORM
             # Fallback and ask the client. We have no Idea who this is.
             color = self.chat_widget.client.player_colors.getUserColor(player.id)
 
-        if mentioned:
+        if mentioned and name_of_author_message != self.chat_widget.client.login:
```
Wesmania: I'd rather we used the 'User' class for nick comparison ('me' member in client, represents the local faf user, has a 'player' member that's not None when we're logged in).

src/chat/channel.py
ghost: Maybe I'm missing your point but what are you thinking then ? 

If I'm not wrong, either the user is an IRCPlayer (if only irc) or Player instance (if he can also play games & co).
If it's an ircplayer then its name is in .name, but for player the name is in .login. So that's why I did like this, otherwise I can't grab his username (or will crash since no attribute).

src/chat/channel.py
ghost: I think I did it correctly ? 
Replaced it by 'self.chat_widget.client.me.player.login'

src/chat/channel.py
Wesmania: Using hasattr is a worse idea than using isinstance. Either way, neither of these things should be necessary - we need to check whether the chatter is us, and I think the better way to do it is to check whether his corresponding player is the same player we're logged in as (or, if chatter.player is client.me.player and both are not None).

src/chat/channel.py
Wesmania: It's serviceable (given that with current mess of a code it's annoying to try and pass 'me' in chatter constructor), but check for me.player being None (might be the case if we lose connection to the server, but keep it to chat). Using some local variable for 'me' should make things a bit shorter.

src/chat/channel.py
ghost: Ok I did it as you said but tbh I'm not sure what you were thinking because I'm confused between all these classes and your "if" only works when he's not an irc user (if irc : login_user, a new variable I added, is None) so if a irc pings me I won't see the message in red ( maybe I'm wrong somewhere in the logic )

src/chat/channel.py
ghost: Ok, I added a new variable for it.

Wesmania:
I'll take this one for myself and do some changes - I hope you won't mind if I do some force-pushing to this branch.

PR #907, opened 2017-12-07
Add option to blink the channel when there is a new message (Fixes #904)

### LINKED ISSUE #904: Notify when messages are sent in non-aeolus channels
As the title states, with the new newbie channel especially this would be very helpful.

ghost:
Add an option in the menu to disable or enable the channel blinking when there is a new message in a channel except for aeolus (it should be on the first time, not sure if that's the case since I can't really test this scenario when it is the first time ever). It also show the icon in the taskbar yellow.
(Did this one because with the new newbie channel, it would be nice for users to know when there are new messages)

And it only blinks if it is the channel you are not currently looking (if you are in #anychannel and someone writes a message in #anychannel, nothing happens).

No blinking for #aeolus even if you don't look aeolus because otherwise it will always blink.

Initial PR:

- [ ] Code has tests

- [ ] Rebase onto develop
- [x] Add changelog entry

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,8 @@
  * Add a button to see the ladder maps pool from play tab (#882)
  * Add 'view in leaderboard' to chat user context menu (#893)
  * Add autojoin #newbie channel and the possibility to autojoin other channels (#899)
+ * Add an option where you choose to blink channels when there is a new message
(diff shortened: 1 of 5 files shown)


Wesmania:
Replaced with blinking chat icons. I think it should work well enough to not need configuring.

PR #909, opened 2017-12-11
make python functions lower case (in chat) fix #908

### LINKED ISSUE #908: internal affair - function naming
I see different functions - Qt, python - with the same name, that is confusing.

example: autoJoin in _clientwindow is a QtSignal, while _chatwidget it's a python function.

I would suggest to make python functions (and variabes) lower case, and keep the mixed case to Qt stuff.

GrotheFAF:
This is a suggestion to keep to the python convention (exampled on the chat dir).

Because, I see different functions - Qt, python - with the same name, that is confusing.

example: autoJoin in _clientwindow is a QtSignal, while in _chatwidget it's a python function.

I made the python functions lower case, and kept the mixed case to Qt stuff.

### DIFF
--- a/src/chat/_avatarWidget.py
+++ b/src/chat/_avatarWidget.py
@@ -22,25 +22,25 @@ class PlayerAvatar(QtWidgets.QDialog):
         self.removeButton = QtWidgets.QPushButton("&Remove users")
         self.grid.addWidget(self.removeButton, 1, 0)
 
-        self.removeButton.clicked.connect(self.removeThem)
+        self.removeButton.clicked.connect(self.remove_them)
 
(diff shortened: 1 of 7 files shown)


Wesmania:
I'll go through these renames - I'm mostly worried that some of this stuff could also be in the UI files for classes, and they would have to be renamed as well.

GrotheFAF:
I checked for not in .ui ... but maybe I missed something, so yes better checked twice.
(and if I missed some in .ui .. I would prefer to undo that rename)

GrotheFAF:
coverage decreased -4.2% ... ? how so?

Wesmania:
Hell if I know.

PR #911, opened 2017-12-11
Fix View Aliases command in chat

GrotheFAF:
started with MathieuB8's PR and refined the output.
used package requests to make it work in py2.6
if requests is unwanted the 2 PRs could be mixed to use json
this is a temporary fix because the use of requests block the client

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -13,6 +13,10 @@ from config import Settings
 
 from model.game import GameState
 
+import json
+import logging
+logger = logging.getLogger(__name__)
+
 """
 A chatter is the representation of a person on IRC, in a channel's nick list.
 There are multiple chatters per channel.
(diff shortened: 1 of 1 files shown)


GrotheFAF:
replaced the requests parts and with MathieuB8's json

src/chat/chatter.py
```diff
@@ -369,11 +373,22 @@ def set_color(self):
             color = pcolors.getUserColor(_id, name)
         self.setForeground(QtGui.QColor(color))
 
+    # temp sync api handling with request to fix viewAliases 12.2017
```
Wesmania: This info will be visible with git blame, we don't neeed it here.

src/chat/chatter.py
```diff
@@ -369,11 +373,22 @@ def set_color(self):
...
+    def return_decoded_json(self, link, method):
+        try:
+            with urllib.request.urlopen(link) as response:
+                return json.loads(response.read().decode())
+        except urllib.error.URLError as e:
+            logger.error("{} method error (link : {}) : {}".format(method, link, e.reason))
+            QtWidgets.QMessageBox.about(self.parent, "Aliases", "Error getting info, "
+                                                                "report the issue with logs on the faf forum")
+            return 'err'
```
Wesmania: I think None would be better here, just so we don't confuse an arbitrary string with json.

src/chat/chatter.py
```diff
@@ -369,8 +369,58 @@ def set_color(self):
...
+                        .format(nicks['attributes']['name'], nicks['attributes']['changeTime'][:10])
+            nick_list += '<tr><td>{}</td><td align="right">currently used</td></tr>' \
+                         '</tbody></table><br/>'.format(self.user.name)
+        else:
+            nick_list = 'The name is not currently owned by any player.'
+        return nick_list
+
     def viewAliases(self):
-        QtGui.QDesktopServices.openUrl(QUrl("{}?name={}".format(Settings.get("USER_ALIASES_URL"), self.user.name)))
+        # QtGui.QDesktopServices.openUrl(QUrl("{}?name={}".format(Settings.get("USER_ALIASES_URL"), self.user.name)))
```
Wesmania: Don't leave commented-out code, we can recover it with git if need be.

src/chat/chatter.py
```diff
@@ -369,8 +369,58 @@ def set_color(self):
...
+        return result
+
+    def names_previously_known(self):
+        api_link = 'https://api.faforever.com/data/player?include=names&fields[player]=login,names&' \
+                   'fields[nameRecord]=name,changeTime&&filter[player]=login==' + self.user.name
+        response = requests.get(api_link).json()
+        if response.get('errors'):
+            return ''
+        if response.get('included') is not None and len(response['included']) >= 1:
+            nick_list = 'The player has previously been known as :' \
```
Wesmania: I think it would be better if we did this formatting through qt .ui files, and then just filled in widgets.

src/chat/chatter.py
GrotheFAF: ok, what about a TODO instead?

src/chat/chatter.py
GrotheFAF: None might be more elegant, but in the 2 function using the return "if response == 'err':",
 that had to become "if response is None:"  and that wouldn't be true. (we don't want to lie, or?)

src/chat/chatter.py
GrotheFAF: ok

GrotheFAF:
I'm concerned with the amount of data that is given back to the request.
The requests might can be tightened to less data doing the same job. (I don't speak api ... yet)

Wesmania:
Unless we're talking about megabytes or hundreds of kilobytes, it shouldn't be a problem. Checking player aliases is something a player does relatively rarely compared to all other client-server communication.

src/chat/chatter.py
Wesmania: In that case throw and catch an error, like ValueError or other standard python exception.

src/chat/chatter.py
Wesmania: Sounds good.

src/chat/chatter.py
GrotheFAF: that is not the point of the return and the error is thrown in the lines above it.

GrotheFAF: but your the Boss .. have it your way.

Wesmania: You can do it how you prefer, but please use some dedicated type instead of a string for reporting an error.

GrotheFAF: here it's not a report. it's a very local flag.

Wesmania: I don't understand what you mean. An error is a special condition, it should be reported with some dedicated value or type, not a magic string.

GrotheFAF: You don't see the lines 382-383 (which do the reporting)?
And I already changed it .. so what is this about?
If you want something else, just tell me what you want there.

Wesmania: Oh, I'm really sorry for that, my fault for not checking if it's outdated. You're absolutely right, it's fine right now.

Wesmania:
I did some more work with this, over [here](https://github.com/Wesmania/client/tree/rewrite-View_Alias-with-api). Did some testing, I think it's merge-worthy.

Wesmania:
Some of the API links in this version don't return the right results sometimes, I fixed it in my version. I'll merge my stuff and close this one just as a formality - I credited you both in the commit message.

PR #912, opened 2017-12-13
add ladder map delayed reveal in chat - fix #360

### LINKED ISSUE #360: Ladder map known before the game begins
You can avoid playing maps you don't like on ladder because the client shows you in a lobby and gives the map name before the game is launched, allowing you to quit with no penalty.

GrotheFAF:
this should fix #360 which might be worsened by now the map shown in chat

add ladder_hide during game state=open to hide game.map and game.title (reveals oponent) to the client user.
untestet - if that's enough .. it should be ... but better be sure.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -15,9 +15,11 @@
  * Hotfix race condition between downloading avatars and removing chatters (#957)
  * Don't proceed with game launch if FA path is invalid (#933, #959)
  * Hotfix an irclib exception popping up in some cases (#958)
+ * Delay displaying of user's ladder game info until launch (#360, #912)
 
 Contributors:
  - Wesmania
+ - Grothe
 
 0.16.1-rc.2
 ====
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -73,6 +73,8 @@ class Chatter(QtWidgets.QTableWidgetItem):
         self.mapItem.setFlags(QtCore.Qt.ItemIsEnabled)
         self.mapItem.setTextAlignment(QtCore.Qt.AlignHCenter)
 
+        self._ladder_hide = False
+
         self._user = None
         self._user_player = None
         self._user_game = None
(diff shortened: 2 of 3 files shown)


src/model/game.py
```diff
@@ -33,6 +33,7 @@ class Game(QObject):
     """
     gameUpdated = pyqtSignal(object, object)
     liveReplayAvailable = pyqtSignal(object)
+    ladderMapReveal = pyqtSignal(object)
```
Wesmania: None of this new code is related to actual model of game data (such as 'live replay for the game is available now' modeled by liveReplayAvailable), it's strictly a matter of how it's shown in the UI. No changes in the model class should be necessary.

You can put this code in some separate listener class that will determine if info should be hidden based on a game, and schedule any timers and signals there.

GrotheFAF: So no changes to game.py, the signal should come from gameset.py newLiveGame ?
I try to do it your way .. if I do it my way, you won't like it... so hint me. .. please.

Wesmania: What we want to achieve here is add a logic of 'if we shouldn't display detailed game info right now, let's not do it, but schedule a state update later'. We could encapsulate this logic in some new 'Sensitive game info checker' class.

An instance of it could have a 'check if the game has (recent) sensitive info' method that takes a game object. Then we can add an 'emit a signal when this game's info can be shown' method that also takes a game object (and that can be cancelled or reset at a game update, or when a game is replaced by a different one). These methods can be called in chatter on every game update, for example.

This will put all 'when and how to delay sensitive game info' logic in one continguous chunk of code, which I think is the cleanest solution.

GrotheFAF:
well, I did it (only) in chatter. So done it 'my way'. Let's see where we go from here. (Don't know how to do the class you wrote about.)

In the end it might be enough, to hide the map while game.state = OPENED, cause when game.state = PLAYING the game is already registered with the server and quitting then will be counted as lost, but not sure .. needs more investigation.

Wesmania:
I tried to do some stuff based on your work, put it over [here](https://github.com/Wesmania/client/tree/add-ladder-map-delayed-reveal). I didn't test it with actual ladder games, but we can try writing some unit tests for the new class. Tell me what you think.

GrotheFAF:
Well, this whole thing might be beside the point - when you think about it - the only one who shouldn't see the ladder map revealed is '.me'. 
If a player is starting a ladder match, the map needs only be hidden to him.
Everbody elses ladder maps don't reveal anything to any effect dis/advantage.
(One could think of buddies revealing each others ladder map over voice ... or alike, but let's not go there)
So the hide might be handled in the line of search_ranked ... handle_game_launch ... and cleared by .me being in game.

Wesmania:
I put a line in my version that checks if the user is one of the players playing ladder and only marks the game as having sensitive data if that's the case.
As for using game launch or ladder search functions, I don't think adding a bit more precision is worth the complexity - we'd create an additional dependency between game tab code, game launching code and chat code. The way it happens now, chatters depend (mostly) only on updates from the server, which makes all logic easier to follow.

GrotheFAF:
Where is that line? (I can't see it)

Wesmania:
Hmm, I must've missed it by accident when refactoring. I re-added it, it's in gameinfo.py:52.

GrotheFAF:
mixing of ideas fixed.

Wesmania:
I rebased my branch built on top of yours.

GrotheFAF:
ok, but not here ...

PR #924, opened 2017-12-18
readded xdelta3 extension [FIX]

sir-ragna:
This was necessary to fix the client on my GNU/Linux system. 
Otherwise it gives me an error message that it can't find xdelta3.exe

Was the extension deliberately missing on platforms that don't identify as **win32**?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 0.x.x
 =====
 
-0.17.0
+0.16.0-rc.2
 ====
  * Introduce new object model:
    - Game, Player and IrcUser classes track corresponding info from the server,
--- a/src/fa/updater.py
+++ b/src/fa/updater.py
@@ -706,6 +706,9 @@ class Updater(QtCore.QObject, ConnectionHandler):
             xdelta = os.path.join(fafpath.get_libdir(), "xdelta3.exe")
         else:
             xdelta = "xdelta3"
+            if not shutil.which(xdelta):
+                QtWidgets.QMessageBox.information(None, "xdelta3 not found", 
+                                    "Please install xdelta3 on your system.")
         subprocess.call([xdelta, '-d', '-f', '-s', original, patch, toFile], stdout=subprocess.PIPE)
         shutil.copy(toFile, original)
         os.remove(toFile)
--- a/src/fa/updater.py
+++ b/src/fa/updater.py
@@ -707,8 +707,7 @@ class Updater(QtCore.QObject, ConnectionHandler):
         else:
             xdelta = "xdelta3"
             if not shutil.which(xdelta):
-                QtWidgets.QMessageBox.information(None, "xdelta3 not found", 
-                                    "Please install xdelta3 on your system.")
+                logger.error("%s not found on the system" % xdelta)
         subprocess.call([xdelta, '-d', '-f', '-s', original, patch, toFile], stdout=subprocess.PIPE)
         shutil.copy(toFile, original)
         os.remove(toFile)


Wesmania:
It is deliberate. Most distros have an 'xdelta3' package that provides the xdelta3 binary. Try finding and installing it to see if the problem goes away by its own.

sir-ragna:
Thanks for the advice.

Still I would still prefer some kind of fall-back mechanism.
Would something like this be acceptable?

```python
# applying delta
if sys.platform == 'win32':
    xdelta = os.path.join(fafpath.get_libdir(), "xdelta3.exe")
elif shutil.which('xdelta3'):
    xdelta = "xdelta3"
else:
    xdelta = "xdelta3.exe"
```

Wesmania:
The purpose to looking for xdelta is to use it later. There's no way to run a windows exe on Linux, so any such fallback is of no use. Best advice I can give is to consider xdelta3 a dependency of the FAF client, same way as PyQt5.

sir-ragna:
> There's no way to run a windows exe on Linux

Yes there is. As I said, adding the extension actually fixes it. The`.exe` is automatically run by Wine. The same is true for the Supreme Commander game itself.

We could add xdelta3 to the dependency list and write a warning to the logs if the fall back is triggered.

Wesmania:
Not all distros automatically run exes through wine. Also, running xdelta through wine could incur significant time cost (like updating the wineprefix), blocking the update process for all this time. Hoping that all default wine version we ever encounter (if any, remember that wine is only used by the client to run the game) will run xdelta with no issues is another matter. I'd rather use native xdelta only, especially since it's a common package.

sir-ragna:
I don't really like my installation breaking like this, but I understand your reasoning.

Maybe we can resolve this by informing the user.
Would the following code be acceptable?

```python
if sys.platform == 'win32':
    xdelta = os.path.join(fafpath.get_libdir(), "xdelta3.exe")
elif shutil.which('xdelta3'):
    xdelta = "xdelta3"
else:
    logger.critical("No native xdelta3 found. Please install xdelta3 on your system.")
    QtWidgets.QMessageBox.information(None, "xdelta3 not found", 
                                             "Please install xdelta3 on your system.")
```

Wesmania:
Packages for existing distros like Gentoo and Arch make xdelta a dependency of the client, so the problem doesn't exist there. I think mentioning xdelta as a dependency should be enough (and if we want to make a check, then IMO it'd be best to check return value of command invocation and report whether it went wrong).

sir-ragna:
> Packages for existing distros like Gentoo and Arch make xdelta a dependency of the client, so the problem doesn't exist there. 

I don't care about places where the problem doesn't exist. I care about this specific case I ran into. I run from source, did an update and it broke.

>  I think mentioning xdelta as a dependency should be enough

Adding it to the `requirements.txt` file does fix my problem because that is used to set up the virtual environment. Sadly the builds are broken :(

Wesmania:
xdelta3 is a system package, you should have it in your repositories and it's enough to have it in PATH. Did you try looking for it in your distro's repos?

sir-ragna:
Yes it exists and yes it works with the package from the repo's.

That isn't the point for me however. I want to prevent other people that are running a similar set up, like me from running into the same issue and wasting time on it. Adding it to the requirments.txt seems like the most reasonable thing to do here since it already contains the other dependencies.

Wesmania:
requirements.txt is meant for Python packages, if we add xdelta to it, it will just download a Python wrapper for the binary. Best thing that can be done here, I guess, is checking return value of subprocess.call and reporting to the user if something went wrong.

PR #926, opened 2017-12-19
change time to ISO in View Alaises

GrotheFAF:
ISO time format should be the norm. (full years and lose the seconds)
some formating .. nicer

### DIFF
--- a/src/client/aliasviewer.py
+++ b/src/client/aliasviewer.py
@@ -120,22 +120,20 @@ class AliasFormatter:
         current_times = [t for t in times if t['time'] is None]
 
         past_times.sort(key=lambda t: t['time'])
-        name_format = "<b>{}</b>"
-        past_format = "until {}"
         current_format = "now"
-        past_strings = [(name_format.format(e['name']),
-                        past_format.format(time.strftime('%X %x', e['time'])))
+        past_strings = [("{}".format(e['name']),
+                         "{}".format(time.strftime('%Y-%m-%d &nbsp; %H:%M', e['time'])))
                         for e in past_times]
-        current_strings = [(name_format.format(e['name']),
+        current_strings = [("{}".format(e['name']),
                            current_format)
                            for e in current_times]
         return past_strings + current_strings
 
     def nick_time_table(self, nicks):
-        table = '<br/><table border="0" cellpadding="0" cellspacing="0" width="250"><tbody>' \
+        table = '<br/><table border="0" cellpadding="0" cellspacing="1" width="230"><tbody>' \
                 '{}' \
                 '</tbody></table>'
-        head = '<tr><th align="left">Name</th><th align="right">Used</th></tr>'
+        head = '<tr><th align="left"> Name</th><th align="right">used until </th></tr>'
         line_fmt = '<tr><td>{}</td><td align="right">{}</td></tr>'
         lines = [line_fmt.format(*n) for n in nicks]
         return table.format(head + "".join(lines))
@@ -172,6 +170,7 @@ class AliasWindow:
     def view_aliases(self, name, id_=None):
         player_aliases = None
         other_users = None
+        user = None
         try:
             other_users = self._api.name_used_by_others(name)
             if id_ is None:
@@ -179,6 +178,7 @@ class AliasWindow:
                 if len(users_now) > 0:
                     id_ = users_now[0]['id']
             if id_ is not None:
+                user = name
                 player_aliases = self._api.names_previously_known(id_)
         except ApiError as e:
             logger.error(e.reason)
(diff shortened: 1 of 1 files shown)


src/client/aliasviewer.py
```diff
@@ -120,22 +120,22 @@ def nick_times(self, times):
...
                            for e in current_times]
         return past_strings + current_strings
 
     def nick_time_table(self, nicks):
-        table = '<br/><table border="0" cellpadding="0" cellspacing="0" width="250"><tbody>' \
+        table = '<br/><table border="0" cellpadding="0" cellspacing="1" width="220"><tbody>' \
                 '{}' \
                 '</tbody></table>'
-        head = '<tr><th align="left">Name</th><th align="right">Used</th></tr>'
+        head = '<tr><th align="left"> Name</th><th align="center"> used until</th></tr>'
```
Wesmania: Make it "Used until" and it goes in. Spaces before headings are unnecessary too, I think.

src/client/aliasviewer.py
```diff
@@ -120,22 +120,22 @@ def nick_times(self, times):
...
+                        past_format.format(time.strftime('%Y-%m-%d &nbsp; %H:%M', e['time'])))
                         for e in past_times]
         current_strings = [(name_format.format(e['name']),
                            current_format)
                            for e in current_times]
         return past_strings + current_strings
 
     def nick_time_table(self, nicks):
-        table = '<br/><table border="0" cellpadding="0" cellspacing="0" width="250"><tbody>' \
+        table = '<br/><table border="0" cellpadding="0" cellspacing="1" width="220"><tbody>' \
```
Wesmania: I changed it to 250 since I encountered some people with long aliases, and with 250 the line "The alias wasn't used by anyone else" doesn't hang to the right so awkwardly, I think. What's your opinion?

GrotheFAF: Name and Date are now less wide, so room should be enough.
And I prefer the table to have some space on the right.
But I changed it a bit.

src/client/aliasviewer.py
GrotheFAF: I know ucase is customary, but I don't like it here. (read it like a sentence 'Name [] used until []' and it make the right side heavier) .. but if you insist, I change it.
The spaces are for a small indent, header-feeling.

Wesmania: Sounds good then, keep it if you want.

src/client/aliasviewer.py
Wesmania: Alright, sounds reasonable.

GrotheFAF:
There is another 'problem':
If the name doesn't exist (spelled wrong) it still says "The name has never been used by anyone else."
(The input is case sensitive, might be usefull to mention that in the InputTool)

Wesmania:
Easiest way to fix that would be to check in aliasviewer.py:146 if the list is empty and return the string without the "anyone" else part, before we do the alias filtering.

GrotheFAF:
did the changes and fixed the problem.

src/client/aliasviewer.py
```diff
@@ -146,7 +146,7 @@ def name_used_by_others(self, others, original_user=None):
 
         others = [u for u in others if u['name'] != original_user]
         if len(others) == 0 and original_user is None:
-            return 'The name has never been used.'
+            return 'The name has never been used.<br/><br/>(the search is case sensitive)'
```
Wesmania: IMO it'd be better to add "(case sensitive)" in line 218 (setLabelText), since this formatting is also used by the player context menu. Added bonus of letting the user know right away that it's case sensitive.

GrotheFAF: the case 'user doesn't exist' will not come up from context menu, so ...
but ok ...not too much text?
fafclient-chat-alias-search

Wesmania: It fits, so I think it's good.

src/client/aliasviewer.py
```diff
@@ -120,22 +120,22 @@ def nick_times(self, times):
         current_times = [t for t in times if t['time'] is None]
 
         past_times.sort(key=lambda t: t['time'])
-        name_format = "<b>{}</b>"
-        past_format = "until {}"
+        name_format = "{}"
+        past_format = "{}"
```
Wesmania: While we're at it, these could be removed, since they don't format anything anymore.

GrotheFAF:
done.

src/client/aliasviewer.py
GrotheFAF: I was worried you never ask .. just a sec... ;-)

GrotheFAF: done

Wesmania:
Ended up fastforward-merging it, even though github complained.

GrotheFAF:
So I can delete this branch.

GrotheFAF:
So I can delete this branch. (will look bad on my resume though ;-)

GrotheFAF:
that doesn't look like my last version.

PR #927, opened 2017-12-19
add line chatter numbers and status, country filter to chat

GrotheFAF:
Add separate line with user, filtered, players and irc numbers in chat.
Add country filter (flag click) and status filter (5 status icons) to chat


src/chat/channel.py
```diff
@@ -158,9 +158,10 @@ def filterNicks(self):
             chatter.setVisible(chatter.isFiltered(self.nickFilter.text().lower()))
 
     def updateUserCount(self):
-        count = len(self.chatters)
-        self.nickFilter.setPlaceholderText(str(count) + " users... (type to filter)")
-
+        irc_count = len([c for c in self.chatters if c._player is None])
```
Wesmania: This will go through the whole chatter list every time the chatter count changes. I guess it's fine if it doesn't show performance issues, but maybe we can just pass in and use the playerset object to get the player count.

GrotheFAF: Thought about that - and it's there self._chatterset._playerset in all it's glory - but one still would have to go through the users in that particular channel, either to count the players or the ircs.

Wesmania:

PR #928, opened 2017-12-21
add player/country-list to chat

GrotheFAF:
add a Player/Country list to the right of chatter
add menu-chat toggle and save-option

This is 'fun-Info'  .. 'ah data!'


Wesmania:

PR #931, opened 2017-12-23
spice up the hosted game info

GrotheFAF:
add teams + observer info (behind N/M Player)
color the player+team if enough player and/or team match
add sim mods hint
niced tooltip

### DIFF
--- a/src/model/game.py
+++ b/src/model/game.py
@@ -90,6 +90,13 @@ class Game(QObject):
         self._live_replay_timer.timeout.connect(self._emit_live_replay)
         self.has_live_replay = False
 
+        # Update highlight
+        self._highlight_timer = QTimer()
+        self._highlight_timer.setSingleShot(True)
+        self._highlight_timer.setInterval(250)
+        self._highlight_timer.timeout.connect(self.update)
+        self.highlight = False
+
         self._update(state, launched_at, num_players, max_players, title,
                      host, mapname, map_file_path, teams, featured_mod,
                      featured_mod_versions, sim_mods, password_protected,
@@ -169,6 +176,16 @@ class Game(QObject):
             self.visibility = visibility
 
         self._check_live_replay_timer()
+        self._check_highlight()
+
+    def _check_highlight(self):
+        if self.state != GameState.OPEN or self._highlight_timer.isActive():
+            return
+        if self.highlight:
+            self.highlight = False
+        else:
+            self.highlight = True
+            self._highlight_timer.start()
 
     def _check_live_replay_timer(self):
         if (self.state != GameState.PLAYING or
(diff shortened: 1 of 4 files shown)


GrotheFAF:
There might be a better solution for the highlight, but I couldn't figure one out.
Sugestions welcome.

src/model/game.py
```diff
@@ -90,6 +90,13 @@ def __init__(self,
         self._live_replay_timer.timeout.connect(self._emit_live_replay)
         self.has_live_replay = False
 
+        # Update highlight
```
Wesmania, line 93: Model classes are supposed to represent games / players etc. as themselves, not tied to any UI presentation, so update signals should only be emitted if the game actually changes. I think it would be better to put this timer in gameitem, wherever the game update signals are connected to.

Wesmania (changes requested):
I think it's good, qt blinking stuff seems to be usually implemented this way. It should be moved out of model classes, though.

src/model/game.py
GrotheFAF: I tried GameItem first ... coudn't figure out where to put the rest (aside form the timer in init)
The call stack for gameitem: runFAF .exec_() -> GameItemDelegate.paint
I'm clueless here ... and I really wanne know.

Wesmania: Okay, I see where the difficulty is. After we switched to a typical qt model / view stuff (which does give us nifty things like perfect sorting, filtering and consistency of data), we lost what little explicit control qt gives us in case of QWidgetLists (which isn't much anyway).

The way it basically works is, we implement a Qt model of a list of things, and this model and its things (GameModelItem and GameModel) emit signals when they change (much like our own player / playerset etc. classes). When we give this to a qt view, it does all visual updates by itself, using our own delegate class to actually draw the stuff (GameItemDelegate).

Long story short, the GameModelItem class in gamemodelitem.py is the class that represents 'the UI representation of a game and all the times it changes'. We can put the timer and the 'highlight' attribute there, and emit the 'updated' signal (which in itself means 'update the UI representation').

As for showing the actual blinking, the 'data' object in GameItemDelegate (line 76 for first appearance) is exactly the GameModelItem stored in our model and passed to the delegate. We can read the 'should highlight' attribute from it in line 186.

GrotheFAF: Yes, I thought something was strange. As I understand it (oldschool), GameItem is 'just' a function (used here to prepare and display games in a magic fashion 'model /view stuff'  ;-)
I will try GameModellItem (even thought it works, the current solution seems wastefull to me)
(and I asume you meant L226 not L186 (wrong 'silver') I get data.highlight there, that's fine)

Wesmania: Yeah, pretty much. If we ever need game stuff in the UI to update based on friend stuff, or players, or something else that affects their display, GameModelItem is the place to go.

GrotheFAF: was an easy move to gammodelitem... worked right away .. but I made a git  mess. made a new PR

PR #934, opened 2017-12-27
fix #930 announcement link

### LINKED ISSUE #930: chat message showing friends game does not load live replay
When a friend is playing there is a message in chat with a link to the live replay. Clicking this link fails to load the live replay since the last client update.

GrotheFAF:
url type QtUrl was not converted to string for msg

+ game.state for announcing host was queried PLAYING instead of OPEN
+ some lower case and line break


Wesmania:
I guess duke would frown at bundling the GameState fix with this, but it looks good to me. Changelog + rebase and it goes in.

GrotheFAF:
added changelog

PR #942, opened 2017-12-31
add all friends to announcements

GrotheFAF:
use new friendtracker (Wesmania) to signal friend-game events
stash signals on client start until client has settled, and push them on addChatter

### DIFF
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -272,9 +272,6 @@ class ClientWindow(FormClass, BaseClass):
 
         self.player_colors = PlayerColors(self.me)
 
-        self.game_announcer = GameAnnouncer(self.gameset, self.me,
-                                            self.player_colors, self)
-
         self.power = 0  # current user power
         self.id = 0
         # Initialize the Menu Bar according to settings etc.
@@ -296,22 +293,8 @@ class ClientWindow(FormClass, BaseClass):
         self.modMenu = None
 
         self._alias_window = AliasSearchWindow(self)
-        #self.nFrame = NewsFrame()
-        #self.whatsNewLayout.addWidget(self.nFrame)
-        #self.nFrame.collapse()
 
-        #self.nFrame = NewsFrame()
-        #self.whatsNewLayout.addWidget(self.nFrame)
-
-        #self.nFrame = NewsFrame()
-        #self.whatsNewLayout.addWidget(self.nFrame)
-
-
-        #self.WPApi = WPAPI(self)
-        #self.WPApi.newsDone.connect(self.on_wpapi_done)
-        #self.WPApi.download()
-
-        #self.controlsContainerLayout.setAlignment(self.pageControlFrame, QtCore.Qt.AlignRight)
+        self.game_announcer = GameAnnouncer(self.players, self.me, self.player_colors, self)
 
     @property
     def state(self):
--- a/src/client/gameannouncer.py
+++ b/src/client/gameannouncer.py
@@ -1,61 +1,86 @@
-from PyQt5.QtCore import QTimer
-from model.game import GameState
-
-from fa import maps
+from chat.friendtracker import build_friends_tracker, FriendEvents
+import time
 
 
 class GameAnnouncer:
-    ANNOUNCE_DELAY_SECS = 35
 
-    def __init__(self, gameset, me, colors, client):
-        self._gameset = gameset
+    def __init__(self, playerset, me, colors, client):
         self._me = me
         self._colors = colors
         self._client = client
 
-        self._gameset.newLobby.connect(self._announce_hosting)
-        self._gameset.newLiveReplay.connect(self._announce_replay)
+        self._friends_event_tracker = build_friends_tracker(me, playerset)
+        self._friends_event_tracker.friendEvent.connect(self._friend_event)
 
         self.announce_games = True
         self.announce_replays = True
-        self._delayed_host_list = []
+        self._delayed_event_list = []
+        self.delay_friend_events = True
 
-    def _is_friend_host(self, game):
-        return (game.host_player is not None
-                and self._me.isFriend(game.host_player.id))
+    def _friend_event(self, player, event):
+        if self.delay_friend_events:
+            self._delayed_event_list.append((player, event))
+        else:
+            self._friend_announce(player, event)
 
-    def _announce_hosting(self, game):
-        if not self._is_friend_host(game) or not self.announce_games:
+    def delayed_friend_events(self, player):
+        if not self.delay_friend_events:
             return
-        announce_delay = QTimer()
-        announce_delay.setSingleShot(True)
-        announce_delay.setInterval(self.ANNOUNCE_DELAY_SECS * 1000)
-        announce_delay.timeout.connect(self._delayed_announce_hosting)
-        announce_delay.start()
-        self._delayed_host_list.append((announce_delay, game))
-
-    def _delayed_announce_hosting(self):
-        timer, game = self._delayed_host_list.pop(0)
-
-        if (not self._is_friend_host(game) or
-           not self.announce_games or
-           game.state != GameState.OPEN):
+        if len(self._delayed_event_list) == 0:
+            self.delay_friend_events = False
             return
-        self._announce(game, "hosting")
+        i = 0
+        for event in self._delayed_event_list:
+            if player in event:
+                player, event = self._delayed_event_list.pop(i)
+                self._friend_announce(player, event)
+            i += 1
 
-    def _announce_replay(self, game):
-        if not self._is_friend_host(game) or not self.announce_replays:
+    def _friend_announce(self, player, event):
+        if player.currentGame is None:
             return
-        self._announce(game, "playing live")
+        game = player.currentGame
+        if event == FriendEvents.HOSTING_GAME:
+            if not self.announce_games:  # Menu Option Chat
+                return
+            if game.featured_mod == "ladder1v1":
+                activity = "started"
+            else:
+                activity = "is <font color='GoldenRod'>hosting</font>"
+        elif event == FriendEvents.JOINED_GAME:
+            if not self.announce_games:  # Menu Option Chat
+                return
+            if game.featured_mod == "ladder1v1":
+                activity = "started"
+            else:
+                activity = "joined"
+        elif event == FriendEvents.REPLAY_AVAILABLE:
+            if not self.announce_replays:  # Menu Option Chat
+                return
+            activity = "is playing live"
+        else:  # that shouldn't happen
+            return
 
-    def _announce(self, game, activity):
-        url = game.url(game.host_player.id).toString()
-        url_color = self._colors.getColor("url")
-        mapname = maps.getDisplayName(game.mapname)
-        fmt = 'is {} {}<a style="color:{}" href="{}">{}</a> (on {})'
         if game.featured_mod == "faf":
             modname = ""
         else:
             modname = game.featured_mod + " "
-        msg = fmt.format(activity, modname, url_color, url, game.title, mapname)
-        self._client.forwardLocalBroadcast(game.host, msg)
+        if game.featured_mod != "ladder1v1":
+            player_info = " [{}/{}]".format(game.num_players, game.max_players)
+        else:
+            player_info = ""
+        time_info = ""
+        if game.has_live_replay:
+            time_running = time.time() - game.launched_at
+            if time_running > 6 * 60:  # already running games on client start
+                time_format = '%M:%S' if time_running < 60 * 60 else '%H:%M:%S'
+                time_info = " runs {}"\
+                    .format(time.strftime(time_format, time.gmtime(time_running)))
+        url_color = self._colors.getColor("url")
+        url = game.url(player.id).toString()
+
+        fmt = '{} {}<a style="color:{}" href="{}">{}</a> ' \
+              '(on <font color="GoldenRod">{}</font> {}{})'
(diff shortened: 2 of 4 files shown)


src/client/_clientwindow.py
```diff
@@ -1062,6 +1045,11 @@ def _tabChanged(self, tab, curr, prev):
     def mainTabChanged(self, curr):
         self._tabChanged(self.mainTabs, curr, self._main_tab)
         self._main_tab = curr
+        # the tab change works after client has initialized
+        if not self.init_ready:
+            self.init_ready = self._main_tab == 1  # chatTab
```
Wesmania: I think it's best to just set it to True here, not everyone switches to chat tab at client start and it'll be confusing if announcements only show once you switch to the chat tab.

src/client/gameannouncer.py
```diff
@@ -1,61 +1,60 @@
-from PyQt5.QtCore import QTimer
 from model.game import GameState
-
-from fa import maps
+import chat.friendtracker
```
Wesmania: `from chat.friendtracker import build_friend_tracker` should work too.

src/client/gameannouncer.py
```diff
@@ -1,61 +1,60 @@
...
-        url = game.url(game.host_player.id).toString()
-        url_color = self._colors.getColor("url")
-        mapname = maps.getDisplayName(game.mapname)
-        fmt = 'is {} {}<a style="color:{}" href="{}">{}</a> (on {})'
-        if game.featured_mod == "faf":
-            modname = ""
+        self._delayed_event_list = []
+
+    def _friend_event(self, player):
+        if self._client.init_ready:
```
Wesmania: We could maybe move the init_ready to this class, and set it to true once the delayed_friend_events gets called (and maybe just call it should_queue_events or something like that).

src/client/gameannouncer.py
```diff
@@ -1,61 +1,60 @@
...
+            self._friend_announce(player)
+        else:
+            self._delayed_event_list.append(player)
+
+    def delayed_friend_events(self):
+        while len(self._delayed_event_list) > 0:
+            player = self._delayed_event_list.pop(0)
+            self._friend_announce(player)
+
+    def _friend_announce(self, player):
```
Wesmania: I think catching the 'event type' enum would make processing this much more readable. We can put processing each event type into a separate function, with some sanity checks to make sure things didn't change in the meantime.

src/client/_clientwindow.py
GrotheFAF: Oh right, you mean the popup+sound?  I have that disabled ..forever, so forgot about those

src/client/gameannouncer.py
GrotheFAF: yes, yes it does.

src/client/gameannouncer.py
GrotheFAF: _friend_announce covers changes in the meantime.
but I guess .. I have to do what you say anyway .. else no cookie ... so why argue.

Wesmania: I'm allergic to functions that don't fit on my 25x80 vga console :)

GrotheFAF: That's what I suspected. (cause all your lines before 80)

src/client/gameannouncer.py
GrotheFAF: yes, yes we can. (but not that name)

src/client/gameannouncer.py
GrotheFAF: there is only one line over 80  (and that could be easy broken), and I do now typos and miss stuff 
.. I'm adapting ;-)

src/client/gameannouncer.py
```diff
@@ -1,61 +1,60 @@
...
-        self._announce(game, "hosting")
+    def delayed_friend_events(self):
+        self.delay_friend_events = False
+        while len(self._delayed_event_list) > 0:
+            player, event = self._delayed_event_list.pop(0)
+            self._friend_announce(player, event)
 
-    def _announce_replay(self, game):
-        if not self._is_friend_host(game) or not self.announce_replays:
+    def _friend_announce(self, player, event):
```
Wesmania: I don't really like how we don't use the event type here at all - it means that the logic for figuring out what event happened to a player is in two places, here and in friend tracker. If we change something, we'll have to change code in two places. Can we instead use the event type, and only do some basic sanity checks (like if the game is open)?

src/client/gameannouncer.py
```diff
@@ -1,61 +1,60 @@
...
+                activity = "is <font color='GoldenRod'>hosting</font>"
+            else:
+                activity = "joined"
+            player_info = " [{}/{}]".format(game.num_players, game.max_players)
+        elif game.state == GameState.PLAYING:
+            if not self.announce_replays:  # Menu Option Chat
+                return
+            activity = "is playing live"
+        else:  # GameSate.CLOSED
+            activity = "<font color='Red'>has left the building</font>"
```
Wesmania: This *will* confuse people. Do we even need an event for a friend leaving a game?

src/client/_clientwindow.py
```diff
@@ -1060,6 +1043,9 @@ def _tabChanged(self, tab, curr, prev):
     def mainTabChanged(self, curr):
         self._tabChanged(self.mainTabs, curr, self._main_tab)
         self._main_tab = curr
+        # the tab change works after client has kinda initialized
+        if self.game_announcer.delay_friend_events:
+            self.game_announcer.delayed_friend_events()
```
GrotheFAF: Sad, the client start announcements can still get lost, if you click Chat 'to early' while chat is still building up.
So either we find another 'chat is ready for lines'-signal place, or - the ugly way - add an extra delay in/before delayed_friend_events.

src/client/gameannouncer.py
GrotheFAF: We can do that ...

src/client/gameannouncer.py
GrotheFAF: Well, that is the 'shouldn't happen' case, I like to have those cases signaled if it happens (makes a bigger splash than a log entry).
But I can live without it, cause I noticed you prefer coding without humor.

Wesmania: Humour is nice if it doesn't seep into user experience. You're not the one who will have to deal with confused users on the tech support forum.

src/client/_clientwindow.py
Wesmania: Let's not complicate things just yet. The ideal way really would be to separate chat from widgets - as in make a "chat model" with representations of channels, chatters, lines etc, and have widgets only show changes. We'd then initialize and hook up model stuff earlier and not worry about anything being lost. I did a bit of work towards this, but stalled a bit, see https://github.com/Wesmania/client/tree/chat-rewrite.

GrotheFAF: If it's not working reliably, that's not complicating things, that's just sad, and should be fixed if possible.
We don't know what percentage of users click on the Chat-Tab right away after client start.

Wesmania: It's not obvious to me which point in the code can be considered "chat widget fully loaded". You're welcome to hunt one down.

src/client/gameannouncer.py
GrotheFAF: But that's the way we'll probably hear that there's something unintended/wrong going on.
(and it's an Elvis reference, so hopefully someone will know it's not really faf context.)
If we skip the case, we'll never know, that might be fine too. If you want it gone, say so.

src/client/_clientwindow.py
GrotheFAF: well, I tried that for some time already, and didn't find any better place ... so far.
I might try some more ...

src/client/gameannouncer.py
Wesmania: It's not very unusual for it to happen, since between the event firing and the queue being freed the game could've ended. Either way it's still better to log it rather than bother the user with something a dev wants to know. Since it will happen occasionally anyway, I'd rather see it gone.

GrotheFAF: the game ended case is handled on top. this is different, maybe you look at the updated code... very eventfull.

src/client/_clientwindow.py
GrotheFAF: I added a 2nd holdof in _chadwidget  - on_join, now the announcement is delivered, but in grey (insteat of friend color), so something is still not ready. (the announced player has a name, but is not yet in chatter, chat is still at the 'one man show' level)

GrotheFAF: well, found a solution over addChatter in channel.py.
There might be room for improvement in elegance, suggestions are welcome.

src/client/gameannouncer.py
```diff
@@ -1,61 +1,86 @@
...
-
-        if (not self._is_friend_host(game) or
-           not self.announce_games or
-           game.state != GameState.OPEN):
+        if len(self._delayed_event_list) == 0:
+            self.delay_friend_events = False
             return
-        self._announce(game, "hosting")
+        i = 0
+        for event in self._delayed_event_list:
```
Wesmania, line 33: I don't really like this. We make delaying player messages dependent on the chat widget loading and on IRC info about chatters. This will interfere with everything else that uses this announcer, like notification popups. I'd rather see this reverted to the original approach.

What is the exact reason why the chat widget misses some events? Missing chat tabs might be one. If we additionally queue these messages in _chatwidget and send them to crucial channels once they load, this should fix our problem.

src/client/gameannouncer.py
```diff
@@ -1,61 +1,86 @@
...
+                time_info = " runs {}"\
+                    .format(time.strftime(time_format, time.gmtime(time_running)))
+        url_color = self._colors.getColor("url")
+        url = game.url(player.id).toString()
+
+        fmt = '{} {}<a style="color:{}" href="{}">{}</a> ' \
+              '(on <font color="GoldenRod">{}</font> {}{})'
+        msg = fmt.format(activity, modname, url_color, url, game.title,
+                         game.mapdisplayname, player_info, time_info)
+        self._client.forwardLocalBroadcast(player.login, msg)
```
Wesmania, line 86: For all I can tell, client's localBroadcast signal is only used by the chat widget. If we give this class its own signal that it emits here and pass the class to the chat widget in the constructor, we can get rid of an unnecessary indirection layer.

src/client/gameannouncer.py
GrotheFAF: Well, that's what the delay is for, waiting for chatwidget and chatter to be ready/there.
Once that is done, that part doesn't play any role anymore.
The chatters are not there. Once a chatter is added his message can be put out.
And queuing the message somewhere else, doesn't help, because 'once they load' is the unknown. Without the chatter added one will end up with a message without a sender/chatter. (I tried that with my 2nd source hold back, didn't work.)

src/client/gameannouncer.py
GrotheFAF: There is an additional use of localBroadcast emit("Scores", message["text"]), in handle_notice in _clientwindow. (but I can't remember ever seen that message, is it in use, does it exist?)
That the GameAnnouncer can signal itself is a nice idea, but it is not a Qt object, so how signaling?

Wesmania: We can add signals to GameAnnouncer exactly like to other objects, we just need to subclass it from QObject.

src/client/gameannouncer.py
Wesmania: For all I can tell, the message ends up in a print_raw method in channel.py. It should be able to handle a sender that's not on the chatter list, if it doesn't, then it should be fixed.

I still have an issue with this code having to do special exceptions just for chat. The chat widget as a whole should listen to these announcements and print them to channels - and if a channel is not up yet, it should queue them by itself and send them all once it's up.

GrotheFAF: Yes, it can handle a message form chatter that isn't there, but the message will get the noplayer/offline color (my 2nd source hold back from above). Is that what we find acceptable. I don't. Of course, we could 'fake' the color, as we know the source of the messages.
And the gameannouncer doesn't  know anything about source of the delayed_friend_events call, it would do that for replays, so no "special exception just for chat". 
But - as I'm sure you insist - let's Qt gameannouncer and move the whole delay line  to chat...
(but there is no 'once it's up' (maybe you can find one, I couldn't) there is still only 'chatter is there'.)

Wesmania: Chat widget has an 'add_channel' method.

GrotheFAF: Doesn't do it. I checked (my 2nd source hold back from above).

Wesmania: It does, but check out line 350 of _chatwidget.py. The tab is added before we hook up the localBroadcast signal. Swapping these would be the quickest fix I imagine, other than the small refactor I described above.

GrotheFAF: What would that change? Would it make the chatter add earlier or faster?
Did you test it? Please do.

Wesmania: You mentioned before that emitting notifications when the channel was added wouldn't lead to them being displayed. I'm just suggesting this might be a reason.

Anyway, the issue of the link being displayed in a wrong color in the chat is rather a fault of the display code - the print_raw function assumed so far that it will only print messages from chatters, which makes no sense when it comes from an announcement.
I suggest we split it into two functions - actual 'print_raw_message' that takes all formatting info like a color and doesn't care about message source, and the old 'print_raw' that gets color info from the chatter as it used to and uses 'print_raw_message' to print it. We can then use the 'print_raw_message' function to print announcements.

PR #945, opened 2017-12-31
fix #943 chat user context menu View Replays in Vault

### LINKED ISSUE #943: fix search Replays in Vault from chat
In chat user kontext menu 'View Replays in Vault' does not set the replay search parameter right.

Function searchVault in _replaywidget.py does a lousy job checking the given parameter.

GrotheFAF:
3 out of 5 given values weren't set. so if there was a prior search it wasn't proper cleared.


Wesmania:
+1, looks good

PR #946, opened 2017-12-31
fix #944 click user game announcement link if game has ended

### LINKED ISSUE #944: fix click user game announcement link if game has ended
click on player live game announcement when game had ended, let to fa starting with error message.

GrotheFAF:
In Chat a click on player live game announcement when game had ended, let to fa starting with error message.
Now it will give a message about game ended, with an option to search the users replays in Vault.
And click on host game announcements will now be checked for game started (message with option to watch the live game) and game ended (message).

### DIFF
--- a/src/chat/channel.py
+++ b/src/chat/channel.py
@@ -1,5 +1,5 @@
 from fa.replay import replay
-
+from model.game import GameState
 import util
 from PyQt5 import QtWidgets, QtCore, QtGui
 import time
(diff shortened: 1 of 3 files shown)


src/chat/channel.py
```diff
@@ -199,7 +199,12 @@ def pingWindow(self):
     def openUrl(self, url):
         logger.debug("Clicked on URL: " + url.toString())
         if url.scheme() == "faflive":
-            replay(url)
+            if replay(url) == "Replays":  # game ended  -> Search Replaya
```
Wesmania: I can't accept this kind of hack. If we want to pull the player id from the url, let's not hack it directly out of the url string, this is fragile as hell. I guess QUrl is the only thing we have here, but making a dedicated class for parsing and constructing these out of player / game data would be much more robust.

src/chat/channel.py
```diff
@@ -199,7 +199,12 @@ def pingWindow(self):
...
+                # hack to get player name from url
+                player_id = int(url.path().split("/")[2].split(".")[0])
+                if player_id in self._chatterset._playerset._players:
+                    player = self._chatterset._playerset._players[player_id].login
```
Wesmania: Holy shit, these two lines. Class members starting with an underscore are private, this should be respected. If we need a playerset in the class, pass it in the constructor. Playerset itself has dict-like semantics, there are plenty examples for it in existing code.

If we make a dedicated class for handling these URLs, then using playerset directly won't even be necessary.

src/fa/replay.py
```diff
@@ -87,7 +87,15 @@ def replay(source, detach=False):
             if url.scheme() == "faflive":
                 mod = QtCore.QUrlQuery(url).queryItemValue("mod")
                 mapname = QtCore.QUrlQuery(url).queryItemValue("map")
-                replay_id = url.path().split("/")[0]
+                replay_id = url.path().split("/")[1]
```
Wesmania: Same comment about the url as above.

src/fa/replay.py
```diff
@@ -87,7 +87,15 @@ def replay(source, detach=False):
...
+                if int(replay_id) not in client.instance.gameset:
+                    if QtWidgets.QMessageBox.question(client.instance, "Live Game ended",
+                                                      "Would you like to look for it in Replays?",
+                                                      QtWidgets.QMessageBox.Yes,
+                                                      QtWidgets.QMessageBox.No) == QtWidgets.QMessageBox.Yes:
+                        return "Replays"
```
Wesmania: Return type doesn't match up.

src/fa/replay.py
```diff
@@ -87,7 +87,15 @@ def replay(source, detach=False):
             if url.scheme() == "faflive":
                 mod = QtCore.QUrlQuery(url).queryItemValue("mod")
                 mapname = QtCore.QUrlQuery(url).queryItemValue("map")
-                replay_id = url.path().split("/")[0]
+                replay_id = url.path().split("/")[1]
+                if int(replay_id) not in client.instance.gameset:
+                    if QtWidgets.QMessageBox.question(client.instance, "Live Game ended",
```
Wesmania: This is blocking, but there's another example of it in this function already, so I guess it's good.

Wesmania (changes requested):

src/chat/channel.py
GrotheFAF: when I added something to the QUrl (it was the game-uid) in the past you ripped my head of, so I worked here with what I got. 
I would have no problem with spicing up the QUrl with 1-2 more attr.
(Keep in mind, this is a fix of a pretty old bug (not a refactor), which wasn't that noticeable with the rare friend host announcement - but will be once the friend announcement will come in.)

src/fa/replay.py
GrotheFAF: aren't we flexible there?

src/chat/channel.py
Wesmania: If I recall correctly, my issue was with the exact same thing - raw parsing of the url string. I'd really like to see a separate class for both building and parsing these game URLs.

src/fa/replay.py
Wesmania: This is not an excuse. This return type means something to the caller, adding something out of the blue is unacceptable.

PR #948, opened 2017-12-31
fix #947 chat country tooltip lcase

### LINKED ISSUE #947: fix chat country tooltip lcase + avatar bug
1. Chatter country tooltip is now lower case, not intended that way.
2. Rarely the chatter vanishes while his Avatar is loaded, leading to a program error.

GrotheFAF:
Chatter country tooltip is currently lower case, not intended that way. fixed that.

### DIFF
--- a/src/chat/chatter.py
+++ b/src/chat/chatter.py
@@ -278,8 +278,8 @@ class Chatter(QtWidgets.QTableWidgetItem):
         if player.country is None or player.country == '':
             country = '__'
         else:
-            country = player.country.lower()
-        self.setIcon(util.THEME.icon("chat/countries/{}.png".format(country)))
+            country = player.country
+        self.setIcon(util.THEME.icon("chat/countries/{}.png".format(country.lower())))
         self.setToolTip(country)
 
     def update_rank(self):


src/chat/chatter.py
```diff
@@ -240,8 +240,11 @@ def updateAvatar(self):
             avatarPix = util.respix(url)
 
             if avatarPix:
-                self.avatarItem.setIcon(QtGui.QIcon(avatarPix))
-                self.avatarItem.setToolTip(self.avatarTip)
+                if isinstance(self, QtWidgets.QTableWidgetItem):
```
Wesmania: I don't understand the purpose of this check.

src/chat/chatter.py
```diff
@@ -271,20 +274,26 @@ def updatePlayer(self):
     def update_country(self):
         player = self.user_player
         if player is None:
+            if self.toolTip() == "":  # EARLY OUT
```
Wesmania: Why is it necessary here? Also doing logic based on the state of the UI is bad practice.

src/chat/chatter.py
```diff
@@ -271,20 +274,26 @@ def updatePlayer(self):
...
             self.setToolTip("")
             return
         # server sends '' for no ip2country-resolution
         if player.country is None or player.country == '':
             country = '__'
         else:
-            country = player.country.lower()
-        self.setIcon(util.THEME.icon("chat/countries/{}.png".format(country)))
+            country = player.country
+        if self.toolTip() == country:  # EARLY OUT
```
Wesmania: I assume it's here to avoid a call to setIcon, but is there a measurable performance impact to not avoiding the call?

src/chat/chatter.py
```diff
@@ -271,20 +274,26 @@ def updatePlayer(self):
...
+            return
+        self.setIcon(util.THEME.icon("chat/countries/{}.png".format(country.lower())))
         self.setToolTip(country)
 
     def update_rank(self):
         player = self.user_player
         if player is None:
+            if self.rankItem.toolTip() == "IRC User":  # EARLY OUT
```
Wesmania: Same as above.

src/chat/chatter.py
```diff
@@ -305,15 +314,16 @@ def update_rank(self):
                                                      tooltip_str)
         else:
             icon_str = "newplayer"
+        if self.rankItem.toolTip() == tooltip_str:  # EARLY OUT
```
Wesmania: Same as above.

src/chat/chatter.py
```diff
@@ -336,12 +346,11 @@ def update_status_tooltip(self):
                                          game.num_players, game.max_players, game.uid)
         else:  # game is None or closed
             game_str = "Idle"
-
+        if self.statusItem.toolTip() == game_str:  # EARLY OUT
```
Wesmania: As above.

src/chat/chatter.py
```diff
@@ -364,6 +373,8 @@ def update_map(self):
         # Map icon handling - if we're in game, show the map if toggled on
         game = self.user_game
         if game is not None and not game.closed() and util.settings.value("chat/chatmaps", False):
+            if self.mapItem.toolTip() == game.mapdisplayname:  # EARLY OUT
```
Wesmania: As above.

src/chat/chatter.py
```diff
@@ -373,6 +384,8 @@ def update_map(self):
 
             self.mapItem.setToolTip(game.mapdisplayname)
         else:
+            if self.mapItem.toolTip() == "":  # EARLY OUT
```
Wesmania: As above.

src/chat/chatter.py
```diff
@@ -305,15 +314,16 @@ def update_rank(self):
...
             icon_str = "newplayer"
+        if self.rankItem.toolTip() == tooltip_str:  # EARLY OUT
+            return
         self.rankItem.setIcon(util.THEME.icon("chat/rank/{}.png".format(icon_str)))
         self.rankItem.setToolTip(tooltip_str)
 
     def updateGame(self):
-        self.update_status_tooltip()
-        self.update_status_icon()
+        self.update_status()
```
Wesmania: No need to fold these two into one if early returning is the only reason.

Wesmania (changes requested):
If we want to avoid resetting icons and tooltips at update (and I don't think it noticeably impacts performance), then we can take advantage of the fact that the player update signal gives us 2 objects - current and previous player state. We can compare these at the start to see if anything meaningful changed.

src/chat/chatter.py
GrotheFAF: may I quote myself: "Rarely the chatter vanishes while his Avatar is loaded, leading to a program error.".
if self is gone the check would result in False, and avoid the the error.

src/chat/chatter.py
GrotheFAF: may I quote myself: "speedup (50% faster on average)"

src/chat/chatter.py
GrotheFAF: I don't understand the logic of that sentence.

src/chat/chatter.py
Wesmania: This makes absolutely no sense though, Chatter is explicitly a subclass of this Qt class. The root issue is that the C++ Qt object underneath is collected, and if this check fails, then it's a (maybe undocumented) side effect. Doing it here is a hack in itself (since the way downloader callbacks work is flawed in itself), but it's okay, as long as you show me somewhere in PyQt docs that this is a proper way of checking if the Qt object underneath was deallocated.

src/chat/chatter.py
Wesmania: 50% is relative. Is updating chatters a performance bottleneck in any way? I also commented on another way of doing that, at the top in the main review comment.

src/chat/chatter.py
Wesmania: You combined these 2 functions into one, probably so that early returning skips the second function too. I don't think it's a valid reason, this could very well be done using return values or exceptions.

GrotheFAF:
all speedup gone and the avatarItem error handled differently.

src/chat/chatter.py
Wesmania: This still seems unnecessary.

src/chat/chatter.py
```diff
@@ -240,8 +240,11 @@ def updateAvatar(self):
             avatarPix = util.respix(url)
 
             if avatarPix:
-                self.avatarItem.setIcon(QtGui.QIcon(avatarPix))
-                self.avatarItem.setToolTip(self.avatarTip)
+                try:
+                    self.avatarItem.setIcon(QtGui.QIcon(avatarPix))
+                    self.avatarItem.setToolTip(self.avatarTip)
+                except:
```
Wesmania: Catch-alls are bad. Try using QObject's destroyed() signal to remember that the qt object was destroyed, then check for that (add comments why it's needed, too).

Wesmania (changes requested):

src/chat/chatter.py
GrotheFAF: if I knew that I would have used it.
but Pycharm doesn't like:
except QtCore.QObject.destroyed()  [self unfilled]  (<- probably this cause Pycharm ..well)
except QtCore.QObject.destroyed(self)  [wrong type]
so what should it be?

yes, missed one vvv

Wesmania: destroyed() is a signal, not an exception type, RTFM. Connect to the signal in __init__, set some 'we were destroyed underneath' flag in the callback, then check it here.

GrotheFAF:
reduced.

Wesmania (approved):

PR #955, opened 2018-01-01
fix #599 Replay Search map - blank handling

### LINKED ISSUE #599: Replay search in vault by map not working
Searching the replay vault by map name does not produce consistent results.

E.g. search for 'Tag Craftious Maximus' or 'Esgaroths Ruins' never produces any results
Search for 'Seton's Clutch' produces only results for 'Fun Setons'

Search should be better! Searching for 'Tag' should produce any results with 'Tag' characters in them. Should be case insensitive.

GrotheFAF:
Map name with blanks leads to useless or no result.
Replacing blanks with * and all works out fine.
Yes, that easy ...  fix #599

One downside: search for 'Super Map' will also give 'Super Duper Map'.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,7 +3,8 @@
 
 0.17.0
 =====
- * Fixed the error message "failed to get uid" which has bad spacing
+ * Fix the error message "failed to get uid" which has bad spacing
+ * Fix replay search map - automatic replacement of blanks with * (#955, #599)
 
 Contributors:
  - MathieuB8
--- a/src/replays/_replayswidget.py
+++ b/src/replays/_replayswidget.py
@@ -465,13 +465,16 @@ class ReplayVaultWidgetHandler(object):
         if modListIndex is not None:
             w.modList.setCurrentIndex(modListIndex)
 
+        # Map Search helper - the secondary server has a problem with blanks (fix until change to api)
+        map_name = w.mapName.text().replace(" ", "*")
+
         """ search for some replays """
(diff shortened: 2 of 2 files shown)


Wesmania:
Yeah, the side effect is less than ideal. Maybe we can contact some server people on slack to help us figure out why the secondary server hates spaces so much? This appears to be a server bug anyway.

Wesmania:
We should also check if the server is properly escaping received strings - that incorrect "Seton's Clutch" search makes me worry that there's some bad interaction with quotation marks somewhere.

GrotheFAF:
Consider how many maps will be affected by the side effect, not that many. 
I think it's at least a big improvement over the current situation. 
and a Seton's Clutch  search runs fine as Seton's*Cltuch, primo results.
and I checked ' replacement .. not neccessary.
(and the secondary will go down soon .. api is comin', that's the plan right? could I start now?)

Wesmania:
Sounds convincing to me. We'll have to modify it for the api anyway, but it won't hurt to add a comment that it's a hack to make searching useable (just so we don't think it's absolutely necessary for things to work).

GrotheFAF:
That's why I commented it 'Map search 'helper'  ... I will expand on that.

GrotheFAF:
expanded

Wesmania:
This is a user-visible change, add a changelog entry, describe that we're using wildcards for spaces now.

GrotheFAF:
add changelog

PR #956, opened 2018-01-02
spice up the hosted game info

GrotheFAF:
add teams + observer info (behind N/M Player)
color the player+team if enough player and/or team match
add sim mods hint
add short highlight on game update
add host time (as not from server [yet] - hosted games on client start get that time)
niced tooltip - signal installed sim-mods in green

### DIFF
--- a/src/games/gameitem.py
+++ b/src/games/gameitem.py
@@ -1,7 +1,10 @@
 import os
+import time
 import util
 from PyQt5 import QtCore, QtWidgets, QtGui
 from fa import maps
+import modvault
+from downloadManager import IconCallback
 
 
 class GameView(QtCore.QObject):
@@ -43,7 +46,7 @@ class GameItemDelegate(QtWidgets.QStyledItemDelegate):
     TEXT_OFFSET = 10
     TEXT_RIGHT_MARGIN = 5
 
-    TEXT_WIDTH = 250
+    TEXT_WIDTH = 275
     ICON_SIZE = 110
     PADDING = 10
 
@@ -156,16 +159,65 @@ class GameItemFormatter:
         hostid = game.host_player.id if game.host_player is not None else -1
         return self._colors.getUserColor(hostid)
 
+    def _player_color(self, game):
+        if game.num_players < game.max_players:
+            pcolor = "silver"  # 'not enough people'
+        else:
+            pcolor = "limegreen"  # #32CD32
+            teams = self._game_teams(game)
+            if len(teams) > 1:
+                # list of team sizes
+                team_list = []
+                for team in teams:
+                    team_list.append(str(len(team)))
+                if len(set(team_list)) > 1:  # not all teams equal
+                    pcolor = "#CACD32"  # ~yellowgreen
+        return pcolor
+
+    def _teams_obs(self, game):
+        teams = self._game_teams(game)
+        observers = self._game_observers(game)
+        if len(teams) == 0 and len(observers) == 0:
+            return ""
+        s_font = "<font size='-1'>"
+        e_font = "</font>"
+        if len(teams) > 0:
+            team_list = []  # list of team sizes
+            for team in teams:
+                team_list.append(str(len(team)))
+            team_str = " in " + " vs ".join(team_list) + (" +" if len(observers) > 0 else "")
+        else:
+            team_str = ""
+        if len(observers) > 0:
+            team_str += " {} O.".format(len(observers))
+        return s_font + team_str + e_font
+
+    def _host_time(self, data):
+        if data.hosted_at is None:
+            data.hosted_at = time.time()
+        return time.strftime("%H:%M", time.localtime(data.hosted_at))
+
+    def _sim_mods(self, game):
+        if game.sim_mods:
+            return "with sim mods (hover for list)"
+        else:
+            return ""
+
     def text(self, data):
         game = data.game
         formatting = {
-            "color": self._host_color(game),
-            "mapslots": game.max_players,
-            "mapdisplayname": game.mapdisplayname,
+            "bcolor": "white" if data.highlight else "silver",
             "title": game.title,
-            "host": game.host,
+            "mapdisplayname": game.mapdisplayname,
+            "pcolor": self._player_color(game),
             "players": game.num_players,
+            "mapslots": game.max_players,
             "playerstring": "player" if game.num_players == 1 else "players",
+            "teams": self._teams_obs(game),
+            "hcolor": self._host_color(game),
+            "host": game.host,
+            "hosttime": self._host_time(data),
+            "sim_mods": self._sim_mods(game),
             "avgrating": int(game.average_rating)
         }
         if self._featured_mod(game):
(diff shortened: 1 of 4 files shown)


src/games/gamemodelitem.py
```diff
@@ -17,9 +17,26 @@ def __init__(self, game, me):
         self._me = me
         self._me.relationsUpdated.connect(self._check_host_relation_changed)
 
+        # Update highlight
+        self._highlight_timer = QTimer()
+        self._highlight_timer.setSingleShot(True)
+        self._highlight_timer.setInterval(250)
+        self._highlight_timer.timeout.connect(self._game_updated)
```
Wesmania: I'd connect it to a new method in case we ever want to do something more at actual game update.

GrotheFAF: yes, splitting it is better.

src/games/gameitem.py
```diff
@@ -268,13 +341,17 @@ def _player_table_row(self, player, align):
 
     def _country_icon_fmt(self, player):
         icon_path_fmt = os.path.join("chat", "countries", "{}.png")
-        icon_path = icon_path_fmt.format(player.country.lower())
+        if player.country is None or player.country == '':
```
ChrisKitching, line 344: Either the system is quite insanely designed, or at least one of these possibilities shouldn't exist?

I would suggest just enforcing that `country is None` when unknown, since then this if-block can be replaced with `country = (player.country or '__').lower()` or such.

src/games/gameitem.py
```diff
@@ -268,13 +341,17 @@ def _player_table_row(self, player, align):
...
+        else:
+            country = player.country.lower()
+        icon_path = icon_path_fmt.format(country)
         icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
         return "<img src='{}'>".format(icon_abs_path)
 
     def _player_fmt(self, player):
         if player == self._me.player:
-            pformat = "<b><i>{}</b></i>"
+            pformat = "<b><i>{}</i></b>"
```
ChrisKitching, line 354: This seems like the sort of obvious fix that could be its own commit, and merged independently of this giant monolithic commit.

Giant monolithic commits are extremely hard to review. This may be why so many of the long-open PRs on this repository are the ones that are like that :D

src/games/gameitem.py
```diff
@@ -286,30 +363,31 @@ def _observers_tooltip(self, observers):
...
-        return "Observers: " + ", ".join(observer_strings)
+        return obs_table_start + ",</td>".join(observer_strings) + obs_table_end
 
     def _mods_tooltip(self, mods):
         if not mods:
             return ""
-        return "<br/>With: " + "<br/>".join(mods.values())
+        mods_start = "<br/>With Sim Mods:"
+        mods_inst = "<br/><font color='green'><i>Installed:</i><br/>" + \
+                    "<br/>".join(mods[0]) + "</font>" if mods[0] else ""
```
ChrisKitching, line 381: Instead of using `mods[0]` for installed mods and `mods[1]` for not-installed mods, it would be much less insane to use a named dictionary field. So people don't have to stare at this for 30 seconds before they figure out what the hell is going on, as I just did.

You could go one better and make it an actual object, and enforce that it's never `None`. This would allow you to avoid the checks (since `join()` with empty-list yields empty-string).

src/games/gamemodelitem.py
```diff
@@ -15,13 +16,34 @@ def __init__(self, game, me, preview_dler):
...
         self.game = game
         self.game.gameUpdated.connect(self._game_updated)
+        self.hosted_at = None  # put here as long it's not coming from server
         self._me = me
         self._me.relationsUpdated.connect(self._check_host_relation_changed)
         self._preview_dler = preview_dler
         self._preview_dl_request = PreviewDownloadRequest()
         self._preview_dl_request.done.connect(self._at_preview_downloaded)
 
+        # Update highlight
```
ChrisKitching, line 26: The complete lack of comments explaining what this "highlight" logic is trying to achieve is unhelpful.

src/games/gameitem.py
```diff
@@ -209,11 +261,27 @@ def _game_observers(self, game):
         return [game.to_player(name) for name in game.observers
                 if game.is_connected(name)]
 
+    def _game_sim_mods(self, game):
+        if not game.sim_mods:
```
ChrisKitching, line 265: Another example of oddly defensive programming.

What's the contract of the `game` object? It would appear sensible to make it always have the `sim_mods` field, and simply set it to empty-list when there are no mods. Having to check for existence of fields all the time is a really good bug factory.

src/games/gameitem.py
```diff
@@ -209,11 +261,27 @@ def _game_observers(self, game):
...
+            return None
+        inst = modvault.getInstalledMods()
+        uids = [mod.uid for mod in inst]
+        mods = ([], [])
```
ChrisKitching, line 269: Oh god why: we have these things called "objects". :D

src/games/gameitem.py
```diff
@@ -209,11 +261,27 @@ def _game_observers(self, game):
         return [game.to_player(name) for name in game.observers
                 if game.is_connected(name)]
 
+    def _game_sim_mods(self, game):
```
ChrisKitching, line 264: Doccomment? What is this function for?

src/games/gameitem.py
```diff
@@ -209,11 +261,27 @@ def _game_observers(self, game):
...
+        inst = modvault.getInstalledMods()
+        uids = [mod.uid for mod in inst]
+        mods = ([], [])
+        for uid in game.sim_mods:
+            if uid in uids:
+                mods[0].append(game.sim_mods[uid])
+            else:
+                mods[1].append(game.sim_mods[uid])
+            mods[0].sort()
+            mods[1].sort()
```
ChrisKitching, line 276: Why are you sorting both lists after any insertion into either list? Just sort the lists at the end, surely?!

src/games/gameitem.py
```diff
@@ -209,11 +261,27 @@ def _game_observers(self, game):
         return [game.to_player(name) for name in game.observers
                 if game.is_connected(name)]
 
+    def _game_sim_mods(self, game):
+        if not game.sim_mods:
+            return None
+        inst = modvault.getInstalledMods()
+        uids = [mod.uid for mod in inst]
+        mods = ([], [])
+        for uid in game.sim_mods:
```
ChrisKitching, line 270: Loop can be replaced with two list comprehensions.

src/games/gamemodelitem.py
```diff
@@ -15,13 +16,34 @@ def __init__(self, game, me, preview_dler):
 
         self.game = game
         self.game.gameUpdated.connect(self._game_updated)
+        self.hosted_at = None  # put here as long it's not coming from server
```
ChrisKitching, line 19: Unclear why you're proposing to add code for a feature that definitely won't work. Maybe add that part after filing the corresponding server PR?

Wesmania:
Replaced with jinja.

PR #961, opened 2018-01-08
refactor mod vault

GrotheFAF:
change modvault into QListView (analog to Game)
separate functions for filter and server-search
add an explain text for the rather odd behavior
reorder some imports

remarks:
1. if the icon fetching in modmodelitem is ok (separate func?), we could loose the complete preview part in moditem.
2. a signal is needed to call update_info after a View update. not sure how.

### DIFF
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -26,6 +26,9 @@ import json
 from model.gameset import Gameset
 from model.player import Player
 from model.playerset import Playerset
+from model.modset import Modset
+from modvault.modmodel import ModModel
+from modvault.moditem import ModViewBuilder
 from modvault.utils import MODFOLDER
 import notifications as ns
 import replays
@@ -152,6 +155,7 @@ class ClientWindow(FormClass, BaseClass):
         self.players = Playerset()  # Players known to the client
 
         self.gameset = Gameset(self.players)
+        self.modset = Modset()
         fa.instance.gameset = self.gameset  # FIXME (needed fa/game_process L81 for self.game = self.gameset[uid])
 
         # Handy reference to the User object representing the logged-in user.
@@ -159,8 +163,8 @@ class ClientWindow(FormClass, BaseClass):
 
         # Qt model for displaying active games.
         self.game_model = GameModel(self.me, self.gameset)
-
-        self.lobby_info = LobbyInfo(self.lobby_dispatch, self.gameset, self.players)
+        self.mod_model = ModModel(self.me, self.modset)
+        self.lobby_info = LobbyInfo(self.lobby_dispatch, self.gameset, self.players, self.modset)
         self.gameset.newGame.connect(self.fill_in_session_info)
 
         self.lobby_dispatch["session"] = self.handle_session
@@ -538,7 +542,9 @@ class ClientWindow(FormClass, BaseClass):
         self.replays = replays.ReplaysWidget(self, self.lobby_dispatch,
                                              self.gameset, self.players)
         self.mapvault = vault.MapVault(self)
-        self.modvault = modvault.ModVault(self)
+        self.modview_builder = ModViewBuilder(self.downloader)
+        self.modvault = modvault.ModVault(self, self.mod_model,
+                                          self.modview_builder)
         self.notificationSystem = ns.Notifications(self, self.gameset,
                                                    self.players, self.me)
 
--- a/src/modvault/moditem.py
+++ b/src/modvault/moditem.py
@@ -1,0 +1,214 @@
+from PyQt5 import QtCore, QtWidgets, QtGui
+from downloadManager import IconCallback
+import os
+import util
+import urllib.parse
+
+
+class ModView(QtCore.QObject):
+    """
+    Helps with displaying mods in the mod widget. Handles updates to view
+    unrelated to underlying data, like downloading mod previews. Forwards
+    interaction with the view.
+    """
+    mod_double_clicked = QtCore.pyqtSignal(object)
+
+    def __init__(self, model, view, delegate, dler):
+        QtCore.QObject.__init__(self)
+        self._model = model
+        self._view = view
+        self._delegate = delegate
+        self._dler = dler
+
+        self._view.setModel(self._model)
+        self._view.setItemDelegate(self._delegate)
+        self._delegate.mod_preview_missing.connect(self.download_mod_preview)
+        self._view.doubleClicked.connect(self._mod_double_clicked)
+        self._view.viewport().installEventFilter(self._delegate.tooltip_filter)
+
+    def download_mod_preview(self, mod):
+        name = os.path.basename(urllib.parse.unquote(mod.thumbnail))
+        cb = IconCallback(name, self._mod_preview_downloaded)
+        self._dler.downloadModPreview(mod.thumbnail, name, cb)  # (url, name, requester)
+
+    # TODO make it a utility function?
+    def _model_items(self):
+        model = self._model
+        for i in range(model.rowCount(QtCore.QModelIndex())):
+            yield model.index(i, 0)
+
+    def _mod_preview_downloaded(self, modname, icon):
+        for idx in self._model_items():
+            mod = idx.data().mod
+            if mod.thumbnail != "":
+                name = os.path.basename(urllib.parse.unquote(mod.thumbnail))
+                if name.lower() == modname.lower():  # Previews are not case-preserving
+                    self._view.update(idx)
+
+    def _mod_double_clicked(self, idx):
+        self.mod_double_clicked.emit(idx.data().mod)
+
+
+class ModItemDelegate(QtWidgets.QStyledItemDelegate):
+    mod_preview_missing = QtCore.pyqtSignal(object)
+    painting = QtCore.pyqtSignal()
+
+    ICON_RECT = 100
+    ICON_CLIP_TOP_LEFT = 3
+    ICON_CLIP_BOTTOM_RIGHT = -7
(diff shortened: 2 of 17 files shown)


src/client/_clientwindow.py
```diff
@@ -169,15 +173,16 @@ def __init__(self, *args, **kwargs):
         self.players = Playerset()  # Players known to the client
 
         self.gameset = Gameset(self.players)
-        fa.instance.gameset = self.gameset  # FIXME
+        self.modset = Modset()
+        fa.instance.gameset = self.gameset  # FIXME(needed fa/game_process L81 for self.game = self.gameset[uid])
```
Wesmania: I wonder if it's me that put this here. Either way it's a really bad idea. If it's me, I'll look into why it's actually here.

src/modvault/moditem.py
```diff
@@ -0,0 +1,211 @@
...
+        self._view.doubleClicked.connect(self._mod_double_clicked)
+        self._view.viewport().installEventFilter(self._delegate.tooltip_filter)
+
+    def download_mod_preview(self, mod):
+        name = os.path.basename(urllib.parse.unquote(mod.thumbnail))
+        cb = IconCallback(name, self._mod_preview_downloaded)
+        self._dler.downloadModPreview(mod.thumbnail, name, cb)  # (url, name, requester)
+
+    # TODO make it a utility function?
+    def _model_items(self):
```
Wesmania, line 35: IIRC we don't actually use this, so it can go. Something to be fixed in GameView :/

src/modvault/moditem.py
```diff
@@ -0,0 +1,211 @@
...
+            if mod.thumbnail != "":
+                name = os.path.basename(urllib.parse.unquote(mod.thumbnail))
+                if name.lower() == modname.lower():  # Previews are not case-preserving
+                    self._view.update(idx)
+
+    def _mod_double_clicked(self, idx):
+        self.mod_double_clicked.emit(idx.data().mod)
+
+
+class ModItemDelegate(QtWidgets.QStyledItemDelegate):
```
Wesmania, line 52: That's a bit of duplication of games widget code, but I guess we'll figure out a way to factor out common parts in the future.

src/modvault/modmodel.py
```diff
@@ -0,0 +1,186 @@
+from PyQt5.QtCore import QAbstractListModel, Qt, QSortFilterProxyModel, QModelIndex
+from .modmodelitem import ModModelItem
+from enum import Enum
+
+
+class ModModel(QAbstractListModel):
```
Wesmania, line 6: Same here, a lil bit of duplication. Guess it's a bit harder to fix here since add / remove signals have different names in different set classes - we'll have to plan out well how to deduplicate the stuff in the future.

src/modvault/modmodel.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(self, model):
+        self._filter_type = self.FilterType.ALL
+        self._search_str = ""
+        ModSortModel.__init__(self, model)
+
+    def filter_accepts_mod(self, mod):
+        search_str = self._search_str
+        if search_str != "":
+            if not (mod.author.lower().find(search_str) != -1 or mod.name.lower().find(search_str) != -1 or
+                    mod.description.lower().find(" " + search_str + " ") != -1):
```
Wesmania: This will fail to find a word if it's at the beginning or end of the description, or if there's punctuation around it. Using the 're' module and the word boundary character '\b' should be robust enough.

Wesmania (commented):
That's a massive single commit to review. I skimmed through it as much as I could and caught some stuff, but it's really hard to review commits that large. I'll probably revisit it later. Please try to split them up into more manageable logical chunks as you go in the future (200 or so lines should be good for the largest ones, better to oscillate around 100, 50, sometimes less).

There's some code duplication between this and the game model code, which is nasty - but the code is much better organized now, so I guess we'll work on deduplicating this in the future. Having a consistent interface for sets (like uniform names for some shared signals, some subclassing) should help.

Mod and modlist classes are simple so I guess I won't demand tests. If we find a way to make some abstract bases for these classes, we might not even need them.

src/client/_clientwindow.py
GrotheFAF: My comment mentioned where it is needed, and it's the only place (I checked).
It could be replaced with an import client in fa/game_process.py and then use client.instance.gameset[uid] in L81(I tested it, works).

src/modvault/moditem.py
GrotheFAF: So remove the whole Preview part, will do.

src/modvault/moditem.py
GrotheFAF: well, I started with copying the game widget code and transformed it into mod widget code.
(That's the way I learn how things work.)

src/modvault/modmodel.py
GrotheFAF: I didn't touch or look in to that. I'll have a look.

GrotheFAF: We could add more to include words at begining and end, or make it easy and remove both " ".

src/modvault/moditem.py
Wesmania: Wait, I meant just the function. Maybe I confused it with some other one, sorry :/ If it's used anywhere, leave it.

src/modvault/modmodel.py
Wesmania: It might actually be better, since people might want to search by parts of words.

GrotheFAF: I guess the restriction to a word for search in description was, that a search in lengthy description would lead to too many unwanted search results. But let's keep it simple .. or maybe have a checkbox 'include description'.

src/modvault/moditem.py
GrotheFAF: As I mentioned in the description (remark 1), the whole preview downloader part is currently unused, because the icon is set in ModModelItem, but without downloading missing icons.
(As you mentioned the preview in gameitem was probably a bad idea, I left this up for debate.)

Wesmania: Alright. I'll just make needed fixes in my downloader refactor branch after we merge this.

GrotheFAF: You fix the downloader refactor, (and after that, I can adapt the downloader into the modvault refactor) and we merge this
or:
We merge this and after that you'll make the needed downloader fixes.
which is it?

GrotheFAF:
There are still some untested details, and also point 2. in description needs a solution.
(the info_update should come after view update, but where is the 'handle' for that.)

src/modvault/modmodel.py
GrotheFAF: (you're aware that you answered my 'X or Y ?' with kinda 'yes' ;-)

src/modvault/moditem.py
Wesmania: Your changes are more localized, so I guess I'll merge my fixes first. Either way we'll have to wait just a little bit (a week or so) for 0.16.1 to stabilize for good.

Wesmania:
I can't find either update_info or info_update by grepping for them, what do you mean?

GrotheFAF:
Line 102 in modvault/__init__.py

GrotheFAF:
I put the signal for update_info right in paint, which works - except when the list turns up empty, hence some more calls after filtering are needed to cover that. (maybe there is an empty_list signal.)

Wesmania:
I get:
```
  File "/mnt/dane/common-home/Pobrane/wes-client/src/modvault/moditem.py", line 79, in paint
    text = self._formatter.text(data)
  File "/mnt/dane/common-home/Pobrane/wes-client/src/modvault/moditem.py", line 184, in text
    "date": QtCore.QDateTime.fromSecsSinceEpoch(mod.date, 0).toString("yyyy-MM-dd"),
AttributeError: type object 'QDateTime' has no attribute 'fromSecsSinceEpoch'
  File "/mnt/dane/common-home/Pobrane/wes-client/src/modvault/moditem.py", line 79, in paint
    text = self._formatter.text(data)
  File "/mnt/dane/common-home/Pobrane/wes-client/src/modvault/moditem.py", line 184, in text
    "date": QtCore.QDateTime.fromSecsSinceEpoch(mod.date, 0).toString("yyyy-MM-dd"),
AttributeError: type object 'QDateTime' has no attribute 'fromSecsSinceEpoch'
```
When opening mod vault tab.

GrotheFAF:
This function was introduced in Qt 5.8, maybe I'm too advanced? What Qt version you're using?
That was fromTime_t() ->  https://doc.qt.io/qt-5.10/qdatetime-obsolete.html

Wesmania:
I'n using PyQt 5.7.1, newer versions are marked as unstable. I think what you're doing here should be doable with regular Python time module.

Linux users will thanks you very much for not forcing them to grab a PyQt version that might not yet be offered by their distro :)

GrotheFAF:
It was Qt  fromTime_t(), and as I couldn't find it in the Qt doc of QDateTime, I thought it was time for a change. But I will change it to some older form.

Wesmania:
"Older" is not a good name, "more supported" is. Time module is something every single Python installation has.

GrotheFAF:
oh, pc-police.  maybe we cold call it "age chalenged" .. still to negativ, better: "of advanced age".
And "as I said" I didn't come up with using QDateTime there, that's why I looked it up.

Wesmania:
Stop spewing confrontational garbage. I'm not taking extra effort just to make fun of you, I'm just making it clear that avoiding extra or too new dependency versions is a good thing.

PR #962, opened 2018-01-08
refactor imports _clientwindow and games

GrotheFAF:
reorder imports in _clientwindows and games/[files]
sort code

### DIFF
--- a/src/client/_clientwindow.py
+++ b/src/client/_clientwindow.py
@@ -1,65 +1,48 @@
-from functools import partial
-
-from PyQt5.QtCore import QUrl, QProcess
-from PyQt5.QtWidgets import QLabel, QMessageBox
+from PyQt5 import QtCore, QtWidgets, QtGui
 
 import config
 import connectivity
-from config import Settings
-import chat
-from model.player import Player
-from model.playerset import Playerset
+from connectivity.helper import ConnectivityHelper
+from client import ClientState, LOBBY_HOST, LOBBY_PORT, LOCAL_REPLAY_PORT
+from client.aliasviewer import AliasSearchWindow
 from client.connection import LobbyInfo, ServerConnection, \
         Dispatcher, ConnectionState, ServerReconnecter
-
-from model.gameset import Gameset
-from games.gamemodel import GameModel
-from games.hostgamewidget import GameLauncher
-from games.gameitem import GameViewBuilder
-
-from client.updater import UpdateChecker, UpdateDialog, UpdateSettings
-from client.update_settings import UpdateSettingsDialog
-from client.theme_menu import ThemeMenu
+from client.gameannouncer import GameAnnouncer
 from client.kick_dialog import KickDialog
+from client.login import LoginWidget
+from client.playercolors import PlayerColors
+from client.theme_menu import ThemeMenu
+from client.updater import UpdateChecker, UpdateDialog
+from client.update_settings import UpdateSettingsDialog
 from client.user import User
 import fa
-from connectivity.helper import ConnectivityHelper
-from fa import GameSession
 from fa.factions import Factions
 from fa.maps import getUserMapsFolder
+from functools import partial
+from games.gamemodel import GameModel
+from games.gameitem import GameViewBuilder
+from games.hostgamewidget import GameLauncher
+import json
+from model.gameset import Gameset
+from model.player import Player
+from model.playerset import Playerset
 from modvault.utils import MODFOLDER
+import notifications as ns
+import replays
+import secondaryServer
+import time
+import util
 from ui.status_logo import StatusLogo
-from client.login import LoginWidget
-from client.aliasviewer import AliasSearchWindow
 from ui.busy_widget import BusyWidget
 
-from client.playercolors import PlayerColors
-from client.gameannouncer import GameAnnouncer
 '''
 Created on Dec 1, 2011
 
 @author: thygrrr
 '''
 
-from PyQt5 import QtCore, QtWidgets, QtGui
-from PyQt5.QtWidgets import QInputDialog, QLineEdit
-
-from client import ClientState, LOBBY_HOST, \
-    LOBBY_PORT, LOCAL_REPLAY_PORT
-
 import logging
-
 logger = logging.getLogger(__name__)
-
-import util
-import secondaryServer
-
-import json
-import sys
-import replays
-
-import time
-import notifications as ns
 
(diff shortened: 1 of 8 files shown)


src/client/_clientwindow.py
```diff
@@ -520,71 +502,62 @@ def resizeWidget(self, globalMousePos):
...
 
         # download manager
         self.downloader = downloadManager.downloadManager(self)
 
         self.loadSettings()
 
-        # Initialize chat
-        self.chat = chat.ChatWidget(self, self.players, self.me)
-
-        self.gameview_builder = GameViewBuilder(self.me,
```
Wesmania, line 543: This stuff was done on purpose to conform to Python 80 character width limit. Autopep8 likes it, and it makes vertically split file editing so much easier :)

GrotheFAF: if we really want to go PEP-8 .. way to go.  do we?

Wesmania: I kinda like sticking to the 80 char limit, since vertical file split is a godsend. How about we leave existing code alone for now, and gradually move to 80 char limit as we refactor? It's the most minor of pep8 things anyway.

GrotheFAF: ok, added 80 guideline to my Pycharm and reduced some.
Any idea why AppVeyor is failing?

Wesmania: Travis says "QWidget: Must construct a QApplication before a QWidget", so I imagine reordering imports causes some module to create a widget before a QApplication is created. It's triggered somewhere in tests.

GrotheFAF: How will I find that, Pycharm runs without a problem.

Wesmania: I'm able to reproduce the error by running runtests.py from repo root, so it's probably a good starting point.

GrotheFAF: Got the test runing.
I reversed the fix for: fa.instance.gameset = self.gameset  [_clientwindow, L155].
Seems the test didn't like the 'import client' in game_process.

Wesmania: Did you verify that running games still works?

GrotheFAF: yes

src/client/_clientwindow.py
```diff
@@ -1,66 +1,49 @@
-from functools import partial
-
-from PyQt5.QtCore import QUrl, QProcess
-from PyQt5.QtWidgets import QLabel, QMessageBox
+from PyQt5 import QtCore, QtWidgets, QtGui
 
-import config
+from config import Settings, is_beta, VERSION
```
Wesmania: I'd rather we kept this import as it was, "is_beta" and "VERSION" can be easily confused with local variables.

src/client/_clientwindow.py
```diff
@@ -1,66 +1,49 @@
...
+from client.theme_menu import ThemeMenu
+from client.updater import UpdateChecker, UpdateDialog
+from client.update_settings import UpdateSettingsDialog
 from client.user import User
 import fa
-from connectivity.helper import ConnectivityHelper
-from fa import GameSession
 from fa.factions import Factions
 from fa.maps import getUserMapsFolder
+from functools import partial
```
Wesmania, line 21: Same here, "partial" looks too much like a local variable name.

src/client/_clientwindow.py
```diff
@@ -169,7 +152,7 @@ def __init__(self, *args, **kwargs):
         self.players = Playerset()  # Players known to the client
 
         self.gameset = Gameset(self.players)
-        fa.instance.gameset = self.gameset  # FIXME
+        fa.instance.gameset = self.gameset  # FIXME (needed fa/game_process L81 for self.game = self.gameset[uid])
```
Wesmania, line 155: Line numbers in comments will get obsolete fast. Do we need this comment?

src/games/_gameswidget.py
```diff
@@ -28,20 +28,18 @@ class GamesWidget(FormClass, BaseClass):
...
+    def __init__(self, client, game_model, me, gameview_builder, game_launcher):
         BaseClass.__init__(self)
-
         self.setupUi(self)
 
         self._me = me
         self.client = client
         self.mods = {}
-        self._gameview_builder = gameview_builder
+        #self._gameview_builder = gameview_builder
```
Wesmania: Remove code instead of commenting it out.

src/games/hostgamewidget.py
```diff
@@ -1,5 +1,5 @@
 from PyQt5 import QtCore
-import modvault
+from modvault import utils
```
Wesmania: Same here, I'd rather keep the 'modvault' prefix to avoid confusing 'utils' with a local variable.

src/model/game.py
```diff
@@ -45,23 +45,10 @@ class Game(QObject):
...
-                 host,
-                 mapname,
-                 map_file_path,
-                 teams,
-                 featured_mod,
-                 featured_mod_versions,
-                 sim_mods,
-                 password_protected,
-                 visibility):
+    def __init__(self, playerset, uid, state, launched_at, num_players,
```
Wesmania: Extra lines don't hurt us, and I think previous presentation was more readable.

src/client/_clientwindow.py
GrotheFAF: ok .. but look at 'before' L8, should that go as well?

src/client/_clientwindow.py
GrotheFAF: I only moved that.

src/client/_clientwindow.py
GrotheFAF: Maybe, but it will be 'around that number' until we 'fix' it.
Obviously, you didn't remember what it was good for.

src/games/_gameswidget.py
GrotheFAF: ok. (I left it for a reason, but with this view .. not needed)
(... and you meant like around L300 in _clientwindow?-)

src/games/hostgamewidget.py
GrotheFAF: ok

src/model/game.py
GrotheFAF: speak for yourself, they hurt me. I like to see more.
But if you insist, I'll undo it.

Wesmania:
That's just a reordering / small cleanup, so I won't ask for a changelog entry. Let's merge.

PR #968, opened 2018-01-12
add sort rating chatter

### LINKED ISSUE #965: allow sorting players by rating
As the title states. Being able to sort players by ladder and global rating would be very useful for setting up games.

GrotheFAF:
add sorting row on top of chatter list
add sort options: name, global, ladder
Fix #965


Wesmania:

PR #1052, opened 2018-06-13
make replayserver port dynamic

muellni:
I'm annoyed by the "Port Occupied" message when starting the client a second time. This should fix it.


muellni:
Btw, I think the game used the port 15000 as default port for the replay server. This means that the 'lobby/relay/port' config option was never usable.

Issue #851, opened 2017-09-01
Use jinja as a parser for client strings


Wesmania:
I was touching tooltip formatting code and came up with an idea. Python allows us to define custom formatters - if we implemented one that allows us to express loops, conditionals and member access, we could both simplify the code and make strings vastly more themeable. We could use something like {player.login} for member access, {%if is_me%}...{%else%}...{%endif%} for conditionals and {%for player in players%}...{%endfor%} for loops.

duk3luk3:
You're describing jinja.

And I don't think we should use that unless we really need it.

Wesmania:
Yeah, I wasn't aware of it. Let's make this about using jinja for things then.

PR #1070, opened 2018-07-27
Conversion of GameTooltipFormatter into Jinja2

FlyingThunder:
This is the experimental state of the conversion of the Lobby Tooltip (hover over lobbies in the "games" tab) into the template module Jinja(2).

There might still be possible bugs around that i do not know of yet, so id be glad if i can get some more feedback to this.

Also, i am aware that the template sytax right now isnt the prettiest - ill do that once im sure that everything is working as intended.
I hope i didnt forget something important. Git is a real challenge for me :)
#851

### DIFF
--- a/res/games/gameitem.qthtml
+++ b/res/games/gameitem.qthtml
@@ -1,0 +1,129 @@
+{# Jinja template for the game tooltip #}
+{%- macro playermacro(player, me, iconpath, teamalign) %}
+        {% if player.login is defined %}
+            {% set player_login = player.login %}
+            {% set player_country = player.country | lower %}
+            {% set player_country_path = iconpath~player_country~".png" %}
+            {% set player_global_rating = player.rating_estimate %}
+        {% else %}
+            {% set player_login = player %}
+            {% set player_country_path = "" %}
+            {% set player_global_rating = "???" %}
+        {% endif %}
+
+        {% if me.login is defined %}
+            {% set me_login = me.login %}
+        {% else %}
+            {% set me_login = me %}
+        {% endif %}
+
+        {# This is needed to prevent a new line created on every "-" #}
+        {% if "-" in player_login %}
+            {% set player_login = player_login | replace("-", "&#8209;") %}
+        {% endif %}
+        {% if "-" in me_login %}
+            {% set me_login = me_login | replace("-", "&#8209;") %}
+        {% endif %}
+
+
+        {% if player.clan == me.clan and player.clan is not none and player.clan is defined and player != me %}
+            {% set player_login = "<font color='blue'>"~player_login~"</font>" %}
+        {% elif player_login == me_login %}
+            {% set player_login = "<b><i>"~player_login~"</b></i>" %}
+        {% endif %}
+
+        {% set width = 0 %}
+        <tr>
+        {% if teamalign == "left" -%}
+            <td align='right' width='{{ width }}'>
+                {{ player_login }}
+            </td>
+            <td>
+                <img src='{{ player_country_path }}'>
+            </td>
+            <td align='right' width='10'>
+                ({{ player_global_rating }})
+            </td>
+        {% else %}
+            <td align='left' width='10'>
+                ({{ player_global_rating }})
+            </td>
+            <td>
+                <img src='{{ player_country_path }}'>
+            </td>
+            <td align='left' width={{ width }}>
+                {{ player_login }}
+            </td>
+        {% endif %}
+         </tr>
+{% endmacro -%}
+
+
+{%- macro versus_string(fontsize) %}
+    <td valign='middle'>
+        <font color ='black' size={{ fontsize }}>VS</font>
+    </td>
+{% endmacro -%}
+
+
+{%- macro modtip(mods) %}
+    <b><br/>With:</b><br/>
+    {{"<br />".join(mods)}}
+{% endmacro -%}
+
+
+{%- macro observertooltip(observer) %}
+     {% if observer.login is defined %}
+        {% set observer_country = iconpath~observer.country~".png" %}
+        {% set observer_global_rating = observer.rating_estimate %}
+        {% set observer_login = observer.login %}
+        <br/>{{ observer_login }} <img src='{{ observer_country }}'> ({{ observer_global_rating }})
+    {% else %}
+        <br/>{{ observer }}
+    {% endif %}
+{% endmacro -%}
+
+
+{%- macro TooltipFormatting(teams, mods, observers, me, iconpath) %}
+    <table border="0" cellpadding="0" cellspacing="5">
+    <tbody>
+    <tr>
+        {% for team in teams %}
+            {% if loop.last %}
+                <td><table>
+                    {% for player in team %}
+                        {{ playermacro(player, me, iconpath, teamalign="right") }}
+                    {% endfor %}
+                </table><td>
+            {% else %}
+                <td><table>
+                    {% for player in team %}
+                        {{ playermacro(player, me, iconpath, teamalign="left") }}
+                    {% endfor %}
+                </table><td>
+                {{ versus_string(fontsize='+{}'.format(1 + (((teams[0])|length + (teams[1])|length)/2)|int)) }}
+            {% endif %}
+        {% endfor %}
+    </tr>
+        {% if observers %}
+            <tr>
+            <td>
+                <b>Observers:</b>
+                {% for observer in observers %}
+                    {{ observertooltip(observer) }}
+                {% endfor %}
+            </td>
+            </tr>
+        {% endif %}
+        {% if mods %}
+            <tr>
+                <td>
+                    {{ modtip(mods) }}
+                </td>
+            </tr>
+        {% endif %}
+{% endmacro -%}
+
+{% if teams[0] is defined or observers is defined %}
+    {{ TooltipFormatting(teams, mods, observers, me, iconpath) }}
+{% endif %}
--- a/src/games/gameitem.py
+++ b/src/games/gameitem.py
@@ -3,7 +3,7 @@ import util
 from PyQt5 import QtCore, QtWidgets, QtGui
 from fa import maps
 import html
-
+import jinja2
 
 class GameView(QtCore.QObject):
     """
@@ -218,99 +218,17 @@ class GameItemFormatter:
 
 
 class GameTooltipFormatter:
-    TIP_FORMAT = str(util.THEME.readfile("games/formatters/tool.qthtml"))
 
     def __init__(self, me):
         self._me = me
-
-    def _teams_tooltip(self, teams):
-        versus_string = (
-            "<td valign='middle' height='100%'>"
-            "<font color='black' size='+5'>VS</font>"
-            "</td>")
-
-        def alignment(teams):
-            for i, team in enumerate(teams):
-                if i == 0:
-                    yield 'left', team
-                elif i == len(teams) - 1:
-                    yield 'right', team
-                else:
-                    yield 'middle', team
-
-        team_tables = [self._team_table(team, align)
-                       for align, team in alignment(teams)]
-        return versus_string.join(team_tables)
-
-    def _team_table(self, team, align):
-        team_table_start = "<td><table>"
-        team_table_end = "</table></td>"
-        rows = [self._player_table_row(player, align) for player in team]
-        return team_table_start + "".join(rows) + team_table_end
-
-    def _player_table_row(self, player, align):
-        if isinstance(player, str):
-            country = "<td></td>"
-        else:
-            country = "<td>{country_icon}</td>"
-        pname = ("<td align='{alignment}' valign='middle' width='135'>"
-                 "{player}"
-                 "</td>")
-        order = [pname, country] if align == 'right' else [country, pname]
-        player_row = "<tr>{}{}</tr>".format(*order)
-
-        if isinstance(player, str):
-            return player_row.format(alignment=align, player=player)
-        else:
-            return player_row.format(
-                country_icon=self._country_icon_fmt(player),
-                alignment=align,
-                player=self._player_fmt(player))
-
-    def _country_icon_fmt(self, player):
-        icon_path_fmt = os.path.join("chat", "countries", "{}.png")
-        icon_path = icon_path_fmt.format(player.country.lower())
-        icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
-        return "<img src='{}'>".format(icon_abs_path)
-
-    def _player_fmt(self, player):
-        if player == self._me.player:
-            pformat = "<b><i>{}</b></i>"
-        else:
-            pformat = "{}"
-        player_string = pformat.format(player.login)
-        if player.rating_deviation < 200:   # FIXME: magic number
-            player_string += " ({})".format(player.rating_estimate())
-        return player_string
-
-    def _observers_tooltip(self, observers):
-        if not observers:
-            return ""
-
-        observer_fmt = "{country_icon} {observer}"
-
-        observer_strings = [observer_fmt.format(
-            country_icon=self._country_icon_fmt(observer),
-            observer=observer.login)
-            for observer in observers]
-        return "Observers: " + ", ".join(observer_strings)
-
-    def _mods_tooltip(self, mods):
-        if not mods:
-            return ""
-        return "<br/>With: " + "<br/>".join(mods.values())
+        template_abs_path = os.path.join(util.COMMON_DIR, "games", "gameitem.qthtml")
+        with open(template_abs_path, "r") as templatefile:
+            self._template = jinja2.Template(templatefile.read())
 
     def format(self, teams, observers, mods):
-        teamtip = self._teams_tooltip(teams)
-        obstip = self._observers_tooltip(observers)
-        modtip = self._mods_tooltip(mods)
-
-        if modtip:
-            modtip = "<br/>" + modtip
-
-        return self.TIP_FORMAT.format(teams=teamtip,
-                                      observers=obstip,
-                                      mods=modtip)
+        icon_path = os.path.join("chat", "countries/")
+        icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
+        return self._template.render(teams=teams, mods=mods.values(), observers=observers, me=self._me.player, iconpath=icon_abs_path)
 
 
 class GameViewBuilder:
--- a/src/chat/chatter_model.py
+++ b/src/chat/chatter_model.py
@@ -224,7 +224,7 @@ class ChatterItemFormatter:
         # chr(0xB1) = +-
         formatting = ("Global Rating: {} ({} Games) [{}\xb1{}]\n"
                       "Ladder Rating: {} [{}\xb1{}]")
-        tooltip_str = formatting.format((int(player.rating_estimate())),
+        tooltip_str = formatting.format((int(player.rating_estimate)),
                                         player.number_of_games,
                                         int(player.rating_mean),
                                         int(player.rating_deviation),
--- a/src/model/game.py
+++ b/src/model/game.py
@@ -229,7 +229,7 @@ class Game(ModelItem):
         if not players:
             return 0
         else:
-            return sum([p.rating_estimate() for p in players]) / len(players)
+            return sum([p.rating_estimate for p in players]) / len(players)
 
     @property
     def mapdisplayname(self):
--- a/src/model/player.py
+++ b/src/model/player.py
@@ -64,8 +64,9 @@ class Player(ModelItem):
         Get the conservative estimate of the players global trueskill rating,
         rounded to nearest 100
         """
-        return round((self.rating_estimate()/100))*100
+        return round((self.rating_estimate/100))*100
 
+    @property
     def rating_estimate(self):
         """
         Get the conservative estimate of the players global trueskill rating


res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
+{%- macro playermacro(player,align) %}
+        {% if player.login is not defined %}
+            {% set player_login = player %}
+        {% else %}
+            {% set player_login = player.login %}
+        {% endif %}
```
norraxx: Why there is two structs? Why once there is `player.login` and another time is just `player`?

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+        
+        {% if player.countrypath is not defined %}
+            {% set player_country = "" %}
+        {% else %}
+            {% set player_country = player.countrypath %}
+            {% set player_global_rating = (player.global_rating[0]-(player.global_rating[1]*3))|int %}
+        {% endif %}
```
norraxx: Better practice is:
```
{% set player_country = "" %}
{% if player.countrypath is defined %}
    {% set player_country = player.countrypath %}
    {% set player_global_rating = (player.global_rating[0]-(player.global_rating[1]*3))|int %}
{% endif %}
```

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+            {% set player_login = player %}
+        {% else %}
+            {% set player_login = player.login %}
+        {% endif %}
+        
+        {% if player.countrypath is not defined %}
+            {% set player_country = "" %}
+        {% else %}
+            {% set player_country = player.countrypath %}
+            {% set player_global_rating = (player.global_rating[0]-(player.global_rating[1]*3))|int %}
```
norraxx: At upper part of code you're checking `{% if player.login is not defined %}`, and there you just work with `player.global_rating`. Is it ok?
Have spaces between `-` and `*`, it's better to read.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
+{%- macro playermacro(player,align) %}
+        {% if player.login is not defined %}
```
norraxx: What will happend in `is defined`, when `player.login` will be `None`? Variable is still defined.... but value is missing.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+            {% set alignment = 'left' %}
+            {% if "-" not in player.login %}
+	            {% set width = 0 %}
+	        {% else %}
+	            {% set width = 130 %}
+	        {% endif %}
+        {% else %}
+            {% set alignment = 'right' %}
+	        {% set width = 130 %}
+        {% endif %}
```
norraxx: Extract `width` and `alignment` outside if/else, and reduce `{% if "-" not in player.login %}` to without else.
Indentation.

What's the different between `align` and `alignment`, what's `width` is? Bad name convention

res/games/gameitem.css
norraxx: define `player_global_rating` as `{% set player_global_rating = "???" %}` and then override it.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+                (???)
+            {% endif %}
+            </td>
+        {% else %}
+            <td align='left' valign='middle' width='10'>
+            {% if player_global_rating is defined %}
+                ({{ player_global_rating }})
+            {% else %}
+                (???)
+            {% endif %}
```
norraxx: if `player_global_rating` will be defined upper, if no need to be defined.

norraxx (changes requested):
Code can be simplified, too much logic in templates. Template engine should "display" output, not solve logic.
I'll continue when I'll be back at home.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+
+{%- macro modtip(mods) %}
+    <b><br/>With:</b>
+    {% for mod in mods %}
+        <br/> {{ mod }}
+    {% endfor %}
+{% endmacro -%}
+
+
+{%- macro obstip(observers) %}
```
norraxx: `obstip` is what? Bad naming convention

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+{%- macro versus_string(fontsize) %}
+    <td align='middle' valign='middle' height='100%'>
+    <font color ='black' size={{ fontsize }}>VS</font>
+    </td>
+{% endmacro -%}
+
+
+{%- macro modtip(mods) %}
+    <b><br/>With:</b>
+    {% for mod in mods %}
```
norraxx: `{{"<br />".join(mods)}}` ?

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+
+{%- macro obstip(observers) %}
+    <b>Observers:</b>
+    {% for observer in observers %}
+        <br/>{{ observer.login }} <img src='{{ observer.countrypath }}'> ({{ (observer.global_rating[0]-(observer.global_rating[1]*3))|int }})
+    {% endfor %}
+{% endmacro -%}
+
+
+{%- macro formatting(teams, mods, observers) %}
```
norraxx: What do we format by `formatting`?? Teams, mods, observers? Bad name for macro, that formats game info.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
+{%- macro playermacro(player,align) %}
```
norraxx: `macro playermacro` doesnt say much. Maybe if we combine "player", "table", "information", "row", we can describe what exactly it does.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+	    {% elif teams[0][3] or teams[1][3] %}
+	        {{ versus_string(fontsize='+4') }}
+        {% elif teams[0][2] or teams[1][2] %}
+            {{ versus_string(fontsize='+3') }}
+	    {% elif teams[0][1] or teams[1][1] %}
+	        {{ versus_string(fontsize='+2') }}
+        {% else %}
+            {{ versus_string(fontsize='+1') }}
+        {% endif %}
+    {% endif %}
```
norraxx: This code is unreadable.
1. you check `teams[1]` but not `teams[0]`.
2. if/elif/elif/elif/else are written very poor, and are unreadable - bad indentation
3. i see there algoritm, that can be written in one line.
```
{{ versus_string(fontsize='+{}'.format(1 + max([len(teams[0]), len(teams[1])]) ) }}
```
Maybe you'll need to set max value of teams into variable forward.
4. I don't understand, why `fontsize` has value `+number`. How does it work?
5. all time different checks, once you check if "value" is set (`is (not) defined`), another time you check, if value conversion to boolean is positive `{% if teams[1] %}`.

res/games/gameitem.css
```diff
@@ -0,0 +1,138 @@
...
+        </td>
+	    </tr>
+    {% endif %}
+{% endmacro -%}
+
+
+
+{% if teams[0] is defined %}
+{{ formatting(teams, mods, observers) }}
+{% endif %}
```
norraxx: It would be nice, to have well formatted code, then it'll be much more readable.
```
{% if teams[0] is defined %}
    {{ formatting(teams, mods, observers) }}
{% endif %} 
```
or
```
{% if observers %}
    <tr>
        <td align='left' colspan="3">
            {{ obstip(observers) }}
        </td>
    </tr>
{% endif %}
```

I don't see much shame, in combining formatting html and jinja tags.

src/games/gameitem.py
```diff
@@ -214,103 +214,31 @@ def tooltip(self, data):
         game = data.game
         teams = self._game_teams(game)
         observers = self._game_observers(game)
-        return self._tooltip_formatter.format(teams, observers, game.sim_mods)
+        return self._tooltip_formatter.format(teams, observers, game.sim_mods,)
```
norraxx: ,

src/games/gameitem.py
```diff
@@ -214,103 +214,31 @@ def tooltip(self, data):
...
-
-        if modtip:
-            modtip = "<br/>" + modtip
-
-        return self.TIP_FORMAT.format(teams=teamtip,
-                                      observers=obstip,
-                                      mods=modtip)
+        def get_file_path(player):
+            if player is None or isinstance(player, str):
+                return
```
norraxx: Why do we need that? How can be possible, that player is `None` or `str`? O_o

`Funcitons should do one thing, and should ti it well.` There you have logic, that check proper data, i think, you should send here clear data.

src/games/gameitem.py
```diff
@@ -214,103 +214,31 @@ def tooltip(self, data):
...
-        obstip = self._observers_tooltip(observers)
-        modtip = self._mods_tooltip(mods)
-
-        if modtip:
-            modtip = "<br/>" + modtip
-
-        return self.TIP_FORMAT.format(teams=teamtip,
-                                      observers=obstip,
-                                      mods=modtip)
+        def get_file_path(player):
```
norraxx: Method name and what it does doesn't correspond.

src/games/gameitem.py
```diff
@@ -214,103 +214,31 @@ def tooltip(self, data):
...
+                icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
+                player.countrypath = icon_abs_path
+        for team in teams:
+            for player in team:
+                get_file_path(player)
+        for observer in observers:
+            get_file_path(observer)
+
+        templatedir = os.path.join("games", "gameitem.css")
+        template_abs_path = os.path.join(util.COMMON_DIR, templatedir)
```
norraxx: `template_abs_path = os.path.join(util.COMMON_DIR, "games", "gameitem.css")`

res/games/gameitem.css
norraxx: what `teams[0]` and `teams[1]` mean? What values are inside?

norraxx (changes requested):

res/games/gameitem.css
FlyingThunder: This is for the case when a player appears "offline" to the server (ask the server guys why that happens). In this case, the player info sent to the client is not an instance of the "player" class with rating, login name, id etc, but just a plain string with the player name. No idea why, but its necessary.

res/games/gameitem.css
FlyingThunder: Of course in that case it would get an error, but that case is, as you can tell, impossible x)

res/games/gameitem.css
FlyingThunder: i took the name of the function from the previous python code for understandability - i dont think its bad naming, when working with the code you would know that its about the toolTIP, and it says "observers" right after it. but it wont be a problem to just change it to something longer and clearer i suppose

res/games/gameitem.css
FlyingThunder: dont quite understand this one, why would putting it into an expression make it better?

res/games/gameitem.css
FlyingThunder: again, i took this from the original code, where the final function was called "format" - but ill change it to something like "TooltipFormatting"

res/games/gameitem.css
FlyingThunder: 1) because teams[0] always exists and is kind of obsolete to check, but i need to check if teams[1] exists because i only need the versus string if there are 2 teams
2)/3) yes, packing up messy code was a thing for later, but thanks a lot for already giving me a good replacement :)
4) dont ask me, ask HTML - it was like that before, and it is working with the plus. i dont know if it  also works without it, i didnt question it as it was that way in the old code too
5) yes, that is kind of messy too, but doesnt really change the outcome, as the teams list is never "false" - its either true or undefined (but i agree, it should be uniformed)

res/games/gameitem.css
FlyingThunder: 1) yes, i did that in the template with jinja syntax, but when i tried intendation with nested HTML syntax it looked garbage (to me), i guess thats personal preference

2) the teams in the data the client gets are very rudimentary packed into a list of teams.
it looks roughly like this:
teams = [[player(id,login,global,ladder),player(id,login,global,ladder)][player(id,login,global,ladder)]]
this would be a lobby where there are 2 players in team 1 (teams[0]) and 1 player in team 2 (teams[1]).

src/games/gameitem.py
FlyingThunder: weird, how did that slip in - i didnt even get an exception for that though oO? python being random i guess

src/games/gameitem.py
FlyingThunder: a player can be none when the tooltip is opened before the client fetches the data as far as i know - not 100% sure, this might be obsolete.
if a player is a string then he appears offline to the client (same server issue i mentioned above)

src/games/gameitem.py
FlyingThunder: i suppose you mean it should be clear that it gets the COUNTRY ICON file path? because it does get the file path

src/games/gameitem.py
FlyingThunder: awkward, thanks xD

res/games/gameitem.css
norraxx: Best solution is to rename it in understandable way.

res/games/gameitem.css
norraxx: rename it in better way, don't just copy code, make it better! :-)

res/games/gameitem.css
norraxx: 2. try to get knowledge, how it works, i don't think, that html needs +1 for font size.

src/games/gameitem.py
norraxx: git add -ip

src/games/gameitem.py
norraxx: first of all, it set country icon to player.

res/games/gameitem.css
FlyingThunder: http://www.ftls.org/en/tutorial/HTML/index4.shtml says it with "+",
https://www.w3schools.com/tags/att_font_size.asp says it without. i guess both work the same way, good point

res/games/gameitem.css
FlyingThunder: the align variable is part of the player data - it tells the template if the player is on the left (teams[0]) or on the right (teams[1]). the alignment variable is just used internally to prettify formatting - it tells the template where to put the name, the rating and the icon within the team colum. changed naming to be more obvious

res/games/gameitem.css
FlyingThunder: 3)
changed it to

        {{ versus_string(fontsize='+{}'.format(1 + (((teams[0])|length + (teams[1])|length)/2)|int)) }}

that looks good too

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+        
+        {% set player_country = "" %}
+        {% set player_global_rating = "???" %}
+        {% if player.countrypath is defined %}
+            {% set player_country = player.countrypath %}
+            {% set player_global_rating = (player.global_rating[0]-(player.global_rating[1]*3))|int %}
+        {% endif %}
+
+        {% set width = 130 %}
+        {% if teams[1] is not defined %}
```
Wesmania: Where does `teams` come from? Shouldn't `alignment` be just the same as `teamalign`?

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+            {% set alignment = 'left' %}
+            {% if "-" not in player.login %}
```
Wesmania: I don't understand why we're special-casing logins with '-' in them.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+            {{ player_login }}
+            </td>
+            <td>
+            <img src='{{ player_country }}'>
+            </td>
+            <td align={{ alignment }} valign='middle' width='10'>
+            ({{ player_global_rating }})
+            </td>
+        {% else %}
+            <td align='left' valign='middle' width='10'>
```
Wesmania: Is it supposed to be left align? And why is only leftmost width variable?

res/games/gameitem.css
Wesmania: Agreed, let's remove the check, it will be defined anyway.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+{%- macro modtip(mods) %}
+    <b><br/>With:</b><br/>
+    {{"<br />".join(mods)}}
+{% endmacro -%}
+
+
+{%- macro observertooltip(observers) %}
+    <b>Observers:</b>
+    {% for observer in observers %}
+        <br/>{{ observer.login }} <img src='{{ observer.countrypath }}'> ({{ (observer.global_rating[0]-(observer.global_rating[1]*3))|int }})
```
Wesmania: I think it's a good idea to move the 'global rating - 3 deviations' to python code, since it signifies a general 'rating we'd like to show for players' concept. We can make it a property of Player class called something like 'displayed_rating', then use it here and in other places.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+                {{ playermacro(player,teamalign="left") }}
+            {% endfor %}
+        </table><td>
+        {% else %}
+        <td><table>
+            {% for player in team %}
+                {{ playermacro(player,teamalign="left") }}
+            {% endfor %}
+        </table><td>
+        {{ versus_string(fontsize='+{}'.format(1 + (((teams[0])|length + (teams[1])|length)/2)|int)) }}
```
Wesmania: If we put the versus string in a 'if not loop.last', we can avoid some duplication.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+
+{%- macro TooltipFormatting(teams, mods, observers) %}
+    <table border="0" cellpadding="0" cellspacing="5">
+    <tbody>
+    <tr>
+    {% for team in teams %}
+        {% if loop.last %}
+        <td><table>
+            {% for player in team %}
+                {{ playermacro(player,teamalign="left") }}
```
Wesmania: This loop should be the place where we pass different alignments, depending on `loop.last` and `loop.first`.

src/games/gameitem.py
```diff
@@ -218,99 +218,26 @@ def tooltip(self, data):
...
-        obstip = self._observers_tooltip(observers)
-        modtip = self._mods_tooltip(mods)
-
-        if modtip:
-            modtip = "<br/>" + modtip
-
-        return self.TIP_FORMAT.format(teams=teamtip,
-                                      observers=obstip,
-                                      mods=modtip)
+        def transform_playercountry_to_path(player):
```
Wesmania: IMO it's better to put larger functions directly as class methods. Just slap an underscore in front of the name and we're good.

src/games/gameitem.py
```diff
@@ -218,99 +218,26 @@ def tooltip(self, data):
...
+            if player is None or isinstance(player, str):
+                return
+            else:
+                icon_path_fmt = os.path.join("chat", "countries", "{}.png")
+                icon_path = icon_path_fmt.format(player.country.lower())
+                icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
+                player.countrypath = icon_abs_path
```
Wesmania: Please don't set extra attributes on classes. Only classes' methods should do that, otherwise we get a real mess. Make the `country_to_path` method receive a country name and return a filepath, then make a dict comprehension, something like this:
```
flags = {p.country: self.country_to_filepath(p.country) for p in team for team in teams}
flags.update({p.country: self.country_to_filepath(p.country) for p in observers})
```
Or construct the dict via regular iteration. Then pass these flags as a separate argument to jinja.

src/games/gameitem.py
```diff
@@ -218,99 +218,26 @@ def tooltip(self, data):
...
+        for team in teams:
+            for player in team:
+                transform_playercountry_to_path(player)
+        for observer in observers:
+            transform_playercountry_to_path(observer)
+
+        template_abs_path = os.path.join(util.COMMON_DIR, "games", "gameitem.css")
+        with open(template_abs_path,"r") as templatefile:
+            template = jinja2.Template(templatefile.read())
```
Wesmania: I suggest making this template once, in \_\_init\_\_. No point in reading and compiling the template from scratch every time.

Wesmania (changes requested):
One last thing once other stuff gets fixed - original code also made the logged-in user's name shown in bold. The user's player object could be accessed via the 'me' object as (possibly None) 'player' attribute. It would be nice to restore this.

res/games/gameitem.css
FlyingThunder: i dont understand, can you explain further? and no, alignment is how the player items are aligned *within the team column* - so if the icon, rating etc are bound to left or right side. teamalign is not a HTML formatting alignment, its decided upon this variable where the entire team column goes

res/games/gameitem.css
FlyingThunder: i dont know what you mean again, maybe because the sniplet you commented on is just too small to show what you ment - the "left" here is only for the country icon, it doesnt change a lot between left and right, but it looked better that way when i tested.

and if you mean the width='10': thats only the width that gets assigned to the country icon, i set it to 10 so it doesnt take some standard width but is always as small as possible

res/games/gameitem.css
FlyingThunder: hm, as it is right now, e.g. in a 1v1v1, the team columns are formatted in the same way - but it doesnt look bad imo, but if you think it looks bad, sure

src/games/gameitem.py
FlyingThunder: okay, didnt knew that, so you mean instead of putting it into the player class, i take sth like the login and the country, put it into a dict, pass the country path + login into the template and in the template get the path per-playerlogin? sounds doable for me

res/games/gameitem.css
FlyingThunder: im a little confused now though - why make a new attribute for the "displayed_rating", but not for the "countrypath"? but yea, putting the rating into the class is np

res/games/gameitem.css
FlyingThunder: because for some reason, when a player has "-" in his name, HTML will make a new line after each "-"... and that looks ugly as hell :/

Wesmania: We should find the root cause of that instead of special-casing it. Maybe we're generating invalid html somewhere or not html-escaping things?

res/games/gameitem.css
Wesmania: I don't think I understand - what does teamalign control in the player macro? Order of name & flag? Either way, `alignment` should be passed as an argument, not deduced from global variables. Maybe it's actually always equal to `teamalign`, anyway?

res/games/gameitem.css
Wesmania: It's the rating column in both cases, though. I'm just getting the feeling that 'alignment' and 'teamalign' overlap a bit, and that maybe it'd be better to use a variable like 'position' that can be 'first', 'middle' or 'last', and determine alignment and order from it.

res/games/gameitem.css
Wesmania: Mainly because a 'displayed rating' property is something dependent on the 'player' entity only, and a path to an icon that represents the country is theme specific - it's related to client display, not to the player himself.

src/games/gameitem.py
Wesmania: You won't even need the login - just make a country -> file mapping and access it in jinja like you would in python - something like `dict[player.country]`.

res/games/gameitem.css
FlyingThunder: yes, youre right, alignment was obsolete, i fixed the reason i implemented on another way and didnt notice it, thanks

res/games/gameitem.css
FlyingThunder: it seems like most of those positional html formatters were obsolete - now i only have one variable, which is teamaligns, that just decides the order of icon - rating - login name

res/games/gameitem.css
FlyingThunder: i dont understand - the versus string is in the `else` block of the `if loop.last` check - isnt that pmuch the same?

res/games/gameitem.css
FlyingThunder: i could not find anything to that - but it may not be directly related to HTML after all. and it does not *always* make a new line, but:

when you dont set the width of the column manually, in some cases it will generate a new line when the string is really long. i prevented that by making the column more wide, but when there is only one team, i set the width to 0 - that forces the box to be as small as possible.

and heres the catch: this works with even very long names. no new line created. ONLY when there is a `-` in the name...

Wesmania: Quick googling suggests this, for example:
https://stackoverflow.com/questions/1904612/stop-wrapping-of-hyphenated-text
It might be that there's word wrap enabled by default. Try setting a no-wrap style.

res/games/gameitem.css
Wesmania: You can avoid duplication of the 'for player in team block' if you pull it out of the if-else and leave just the versus-string.

FlyingThunder:
Ah, is it the good standard way to always instead of "If x do y, else z" do
"do y, if not x do z" so you only have one check?

Igor Kotrasiński <notifications@github.com> schrieb am Mi., 1. Aug. 2018
20:42:

res/games/gameitem.css
FlyingThunder: yes, i tried that - it works with normal long names, but still, as soon as there are "-" (now i know theyre called hyphen xD) involved, theres a new line :(

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
+{%- macro playermacro(player, me, iconpath, teamalign) %}
+
+
+
```
Wesmania: No need for so much whitespace.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+        {% set player_login = player.login %}
+        {% set player_country = "" %}
+        {% set player_global_rating = "???" %}
+        {% if player.login is defined %}
```
Wesmania: IMO checking if player is of correct type is a bit more accurate, but this is fine too.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+            {% set player_country = iconpath~player.country~".png" %}
+            {% set player_global_rating = player.rating_estimate %}
+        {% endif %}
+
+        {% set width = 0 %}
+            {% if "-" in player.login %}
```
Wesmania: I'd still rather see a solution that disables word wrap.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+                {% set width = 200 %}
+            {% endif %}
+
+        {% if player_login is not defined %}
+            {% set player_login = player %}
+        {% endif %}
+
+        {% if player.login == me.login %}
```
Wesmania: The player in 'me' might be None, please check jinja docs if it doesn't cause trouble and add extra checks if it does.

res/games/gameitem.css
```diff
@@ -0,0 +1,117 @@
...
+{%- macro modtip(mods) %}
+    <b><br/>With:</b><br/>
+    {{"<br />".join(mods)}}
+{% endmacro -%}
+
+
+{%- macro observertooltip(observers) %}
+    <b>Observers:</b>
+    {% for observer in observers %}
+        <br/>{{ observer.login }} <img src='{{ observer.countrypath }}'> ({{ observer.rating_estimate }})
```
Wesmania: Observers can be of string type as well, so we'll need a check here. Something simple like a bare `{{observer}}` if it's a string.

src/client/_clientwindow.py
```diff
@@ -302,7 +302,7 @@ def __init__(self, *args, **kwargs):
         self.mainTabs.setTabIcon(self.mainTabs.indexOf(self.vaultsTab), util.THEME.icon("client/mods.png"))
         self.mainTabs.setTabIcon(self.mainTabs.indexOf(self.ladderTab), util.THEME.icon("client/ladder.png"))
         self.mainTabs.setTabIcon(self.mainTabs.indexOf(self.tourneyTab), util.THEME.icon("client/tourney.png"))
-        self.mainTabs.setTabIcon(self.mainTabs.indexOf(self.unitdbTab), util.THEME.icon("client/twitch.png"))
+        self.mainTabs.setTabIcon(self.mainTabs.indexOf(self.unitdbTab), util.THEME.icon("client/unitdb.png"))
```
Wesmania: That's part of a different commit, isn't it?

src/games/gameitem.py
```diff
@@ -218,99 +218,18 @@ def tooltip(self, data):
 
 
 class GameTooltipFormatter:
-    TIP_FORMAT = str(util.THEME.readfile("games/formatters/tool.qthtml"))
+
+    icon_path = os.path.join("chat", "countries/")
```
Wesmania: Forward slash might not work on windows, I'm not sure. Using the os-specific separator at the end might be a good idea.

src/games/gameitem.py
```diff
@@ -218,99 +218,18 @@ def tooltip(self, data):
...
-            country_icon=self._country_icon_fmt(observer),
-            observer=observer.login)
-            for observer in observers]
-        return "Observers: " + ", ".join(observer_strings)
-
-    def _mods_tooltip(self, mods):
-        if not mods:
-            return ""
-        return "<br/>With: " + "<br/>".join(mods.values())
+        template_abs_path = os.path.join(util.COMMON_DIR, "games", "gameitem.css")
```
Wesmania: This is something that the theme class should provide us (opening and giving us file contents). I think it's better to use it. Also .css file ending is very unfortunate, we should change it so something like .jinja or .qhtml or the like.

src/games/gameitem.py
```diff
@@ -218,99 +218,18 @@ def tooltip(self, data):
...
+        with open(template_abs_path, "r") as templatefile:
+            self.template = jinja2.Template(templatefile.read())
```
Wesmania: _template would be a nicer, private name.

src/games/gameitem.py
```diff
@@ -218,99 +218,18 @@ def tooltip(self, data):
 
 
 class GameTooltipFormatter:
-    TIP_FORMAT = str(util.THEME.readfile("games/formatters/tool.qthtml"))
+
+    icon_path = os.path.join("chat", "countries/")
+    icon_abs_path = os.path.join(util.COMMON_DIR, icon_path)
```
Wesmania: This stuff is used in just one function, we should move it there. In general class namespace should only contain class-specific constants and stuff like that.

res/games/gameitem.css
FlyingThunder: yes, me too, but i tried everything i could find that HTML gives - nothing worked :/ IMO its better to have a small part of ugly code than an ugly user experience but i can remove it if its important

src/games/gameitem.py
FlyingThunder: it works on my windows

src/games/gameitem.py
FlyingThunder: yes, i was wondering whats up with the .css too - i just took the naming convention from the other jinja template (chat) :D

src/client/_clientwindow.py
FlyingThunder: yes, what the hell. i have no clue why this slipped into my push, it didnt show up when i commited my last changes oO

res/games/gameitem.jinja
```diff
@@ -0,0 +1,120 @@
...
+        {% set player_login = player.login %}
+        {% set player_country = "" %}
+        {% set player_global_rating = "???" %}
+        {% if player.login is defined %}
+            {% set player_country = iconpath~player.country~".png" %}
+            {% set player_global_rating = player.rating_estimate %}
+        {% endif %}
+
+        {% set width = 0 %}
+            {% if "-" in player.login %}
```
Wesmania: Did you try replacing the '-' with a non-breaking hyphen `&#8209;`? Jinja should allow for replacing substrings. Also whatever the solution, we should add a (jinja) comment explaining why it's needed.

res/games/gameitem.jinja
```diff
@@ -0,0 +1,120 @@
...
+        {% if mods %}
+            <tr>
+            <td>
+            {{ modtip(mods) }}
+            </td>
+            </tr>
+        {% endif %}
+{% endmacro -%}
+
+{% if teams[0] is defined or observers is defined %}
```
Wesmania: Do things still break when the team list is empty / None? If so, then we should fix that in the class that prepares the tooltip data.

src/chat/channel_view.py
```diff
@@ -190,7 +190,7 @@ def build(cls, theme, player_colors, **kwargs):
 
     def _load_template(self):
         self._env = jinja2.Environment()
-        template_str = self._theme.readfile("chat/channel.css")
+        template_str = self._theme.readfile("chat/channel.jinja")
```
Wesmania: Does it come with the file rename? Either way I think we should keep the .css suffix (or maybe add a double css.jinja one?)

res/games/gameitem.jinja
```diff
@@ -0,0 +1,120 @@
...
+            {% endif %}
+
+        {% if player_login is not defined %}
+            {% set player_login = player %}
+        {% endif %}
+
+        {% if me.login is defined and player.login == me.login %}
+            {% set player_login = "<b><i>"~player_login~"</b></i>" %}
+
+        {% elif player.clan == me.clan and player.clan is not none %}
```
Wesmania: Missed a 'me' check compared to previous one. Anyway, do jinja docs say something about comparing with undefined values? If they don't compare with anything, we might get away with not checking 'me' here.

FlyingThunder: wym missing? there is a check checking me.login that accounts for all possible anomalies that could happen to me, right?

FlyingThunder: wym missing? there is a check checking me.login that accounts for all possible anomalies that could happen to me, right?

res/games/gameitem.jinja
FlyingThunder: well things dont break because of the checks - will an addition in python change the outcome though? i dont know what id have to fix in the tooltip data class

src/chat/channel_view.py
FlyingThunder: you said the css is unfortunate and that we should change that Q.Q and yes the file is renamed too

res/games/gameitem.jinja
Wesmania: If 'me' is undefined, then you execute the 'elif', so you can end up using undefined 'me' there. Anyway jinja code suggests that undefined values compare falsely with defined values, so as long as we compare 'me' stuff with defined values we should be fine.
As in, compare with player_login in that first loop, and check if player.clan is defined before the comparison.

src/chat/channel_view.py
Wesmania: I mean, .css is unfortunate for the tooltip, since it's not a css file in any way. The channel file is actually a template of a css file.

res/games/gameitem.jinja
Wesmania: Teams and observers are prepared in GameItemFormatter class, in _game_teams and _game_observers functions. Both of these are guaranteed to give us lists as far as I can tell - perhaps the problem lies in mods arguments being None? It does come directly from the game object.

src/chat/channel_view.py
Wesmania: It's unfortunate for the tooltip file, since what's generated there is not a .css file, but an html subset handled by qt. The channel file does generate to css.

res/games/gameitem.jinja
FlyingThunder: Well, clan is set to None in the player.py, so it is always defined, isnt it? i dont really get what i should change about that now, you mean i should put both the login==me and the clan==me check into a bigger me=defined check instead of having the clan==me check as an elif?

src/chat/channel_view.py
FlyingThunder: ah, alright, so qthtml makes sense, i thought itd be a good way to have all jinja templates share one naming, but that makes sense too

res/games/gameitem.jinja
FlyingThunder: worked!

res/games/gameitem.jinja
FlyingThunder: i could not reproduce while at home. it may be related to my slow internet on work after all - the latency is probably very high there.

res/games/gameitem.qthtml
```diff
@@ -0,0 +1,123 @@
...
+        {% set width = 0 %}
+
+        {% if player.login is defined %}
+            {% set player_country = iconpath~player.country~".png" %}
+            {% set player_global_rating = player.rating_estimate %}
+        {% endif %}
+
+        {% if player_login is not defined %}
+            {% set player_login = player %}
+        {% endif %}
```
Wesmania: We can put the above 12 lines into one `{% if player.login is defined %} ... {% else %} ...` just for niceness, and move width down near its use.

res/games/gameitem.qthtml
```diff
@@ -0,0 +1,123 @@
...
+
+
+        {# This is needed to prevent a new line created on every "-" #}
+        {% if "-" in player_login %}
+            {% set player_login = player_login | replace("-", "&#8209;") %}
+        {% endif %}
+
+        {% if me.login is defined and player.login == me.login or
+              me.login is not defined and player == me %}
```
Wesmania: If `me.login` is not defined then `me` is None and the other comparison makes no sense. I suggest just `player_login == me.login` since the earlier is always defined and will compare falsely if `me.login` is undefined.

res/games/gameitem.qthtml
```diff
@@ -0,0 +1,123 @@
...
+            {% set player_login = "<b><i>"~player_login~"</b></i>" %}
+        {% elif player.clan == me.clan and player.clan is not none %}
```
Wesmania: Check if `player.clan` is defined here.

res/games/gameitem.qthtml
```diff
@@ -0,0 +1,123 @@
...
+{%- macro observertooltip(observer) %}
+     {% if observer.login is defined %}
+        {% set observer_country = iconpath~observer.country~".png" %}
+        {% set observer_global_rating = observer.rating_estimate %}
+    {% endif %}
+    {% set observer_login = observer.login %}
+    {% if observer_login is not defined %}
+        {% set observer_login = observer %}
+    {% endif %}
+    <br/>{{ observer_login }} <img src='{{ observer_country }}'> ({{ observer_global_rating }})
```
Wesmania: If observer is just a string, it will create ugly empty parentheses. How about doing a big `{% if observer is str %} ... {% else %} ...` where you either have `<br/>{{observer}}` when it's a string and do the whole logic without checking for undefined stuff otherwise?

res/games/gameitem.qthtml
```diff
@@ -0,0 +1,123 @@
...
+                {{ versus_string(fontsize='+{}'.format(1 + (((teams[0])|length + (teams[1])|length)/2)|int)) }}
+            {% endif %}
+        {% endfor %}
+    </tr>
+        {% if observers %}
+            <tr>
+            <td>
+                <b>Observers:</b>
+                {% for observer in observers %}
+                    <span style='white-space: nowrap;'>{{ observertooltip(observer) }}</span>
```
Wesmania: Leftover from trying to fix hyphen issues, or intentional?

Wesmania (commented):
Everything looks fine, i'd just like to ask for one more feature - original code hides player rating if player deviation is more than 200. Can we have that here?

Issue #740, opened 2017-05-30
Unit database tab has the Twitch icon


JaggedAppliance:
As the title says. I'm not sure what a good replacement icon would be, maybe an acu icon?

It has the twitch icon because it used to be the livestreams tab in case anyone is wondering.

duk3luk3:
That's because someone needs to make the livestreams tab work again.

Wesmania:
We can still make an issue to add a units DB icon though, I'm sure we'll be doing a livestream tab from scratch anyway.

GrotheFAF:
That Unit-DB icon was newly designed by Schmidt&Hirsch-Design Hamburg (and these guys aren't cheap). It symbolizes units on an info-sheet, it only coincidently looks alike the old twitch icon. ;-)

PR #1071, opened 2018-07-30
#740 - Changed the icon of the unitdb tab to something more reasonable than …

FlyingThunder:
#740
…the twitch icon

took a scale/libra/however you call it because unit db kind of represents balance? (icon from https://icons8.de, its free to use for open source projects)

### DIFF
--- a/res/games/gameitem.qthtml
+++ b/res/games/gameitem.qthtml
@@ -1,0 +1,129 @@
+{# Jinja template for the game tooltip #}
+{%- macro playermacro(player, me, iconpath, teamalign) %}
+        {% if player.login is defined %}
+            {% set player_login = player.login %}
+            {% set player_country = player.country | lower %}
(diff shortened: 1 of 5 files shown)


Exotic-Retard:
in my theme which might become default at some point this would need fixing i.e. renaming the icon to the same thing as in that commit so it works. the theme is(?) shipped with the client atm so it needs to be maintained too. (not sure how mazor set it all up or what the plans are even for it.)

Wesmania:
I'll make sure to fix that, once I do we'll merge this. Could you delete the old icon? I don't think it's used anywhere else and we can always restore it.

FlyingThunder:
Ah, i kept it in because the "Streams" Tab might come back eventually, but sure

Wesmania:
The hell, why did this close? Well, I'll merge it in anyway.

PR #1078, opened 2018-11-16
elaborate on UID calc failed warning

Geosearchef:
add a text to the UID error telling users what to do to prevent them from showing up in the forums/technical help channel unless they fail.


Wesmania:
You missed a '"' there.

PR #1085, opened 2019-05-13
map generator

Strogoo:
I doubt this feature will be popular (not like we have tons of good maps in vault, but everyone keeps playing seton/gap, right? :D), w/e. Since player can't join mapgen lobbies without generator, we need this in py client too.

Should wait for java client release and test it, so no need to rush with merging.

### DIFF
--- a/src/fa/check.py
+++ b/src/fa/check.py
@@ -24,7 +24,13 @@ def map_(mapname, force=False, silent=False):
     if fa.maps.isMapAvailable(mapname):
         logger.info("Map is available.")
         return True
-
+ 
+    if "neroxis_map_generator" in mapname:
+        import client # FIXME
+        if not path(client.instance):
+            return False
+        return client.instance.map_generator.generateMap(mapname)
+ 
     if force:
         return fa.maps.downloadMap(mapname, silent=silent)
 
--- a/src/fa/check.py
+++ b/src/fa/check.py
@@ -11,6 +11,8 @@ from fa.mods import checkMods
 from fa.path import writeFAPathLua, validatePath
 from fa.wizards import Wizard
 import util
+from mapGenerator.mapgenUtils import isGeneratedMap
+import client
 
 logger = logging.getLogger(__name__)
 
@@ -24,13 +26,10 @@ def map_(mapname, force=False, silent=False):
     if fa.maps.isMapAvailable(mapname):
         logger.info("Map is available.")
         return True
- 
-    if "neroxis_map_generator" in mapname:
-        import client # FIXME
-        if not path(client.instance):
-            return False
+
+    if isGeneratedMap(mapname):
         return client.instance.map_generator.generateMap(mapname)
- 
+
     if force:
         return fa.maps.downloadMap(mapname, silent=silent)
 
--- a/src/fa/maps.py
+++ b/src/fa/maps.py
@@ -238,7 +238,13 @@ def __exportPreviewFromMap(mapname, positions=None):
         return previews
 
     mapname = os.path.basename(mapdir).lower()
-    mapfilename = os.path.join(mapdir, mapname.split(".")[0]+".scmap")
+    
+    # All these mapgen checks in name here and there
+    # but what can we do? :)
+    if "neroxis_map_generator" in mapname:
+        mapfilename = os.path.join(mapdir, mapname+".scmap")    
+    else:
+        mapfilename = os.path.join(mapdir, mapname.split(".")[0]+".scmap")
 
     mode = os.stat(mapdir)[0]
     if not (mode and stat.S_IWRITE):
--- a/src/fa/maps.py
+++ b/src/fa/maps.py
@@ -20,6 +20,8 @@ import fa
 from config import Settings
 from vault.dialogs import downloadVaultAssetNoMsg
 
+from mapGenerator.mapgenUtils import isGeneratedMap
+
 logger = logging.getLogger(__name__)
 
 route = Settings.get('content/host')
@@ -239,9 +241,7 @@ def __exportPreviewFromMap(mapname, positions=None):
 
     mapname = os.path.basename(mapdir).lower()
     
-    # All these mapgen checks in name here and there
-    # but what can we do? :)
-    if "neroxis_map_generator" in mapname:
+    if isGeneratedMap(mapname):
         mapfilename = os.path.join(mapdir, mapname+".scmap")    
     else:
         mapfilename = os.path.join(mapdir, mapname.split(".")[0]+".scmap")
--- a/src/mapGenerator/mapgenManager.py
+++ b/src/mapGenerator/mapgenManager.py
@@ -1,0 +1,156 @@
+# system imports
+import logging
+import sys
+from PyQt5 import QtCore, QtGui, QtWidgets, QtNetwork
+import os
+import re
+import shutil
+import random
+
+# local imports
+from config import Settings
+import fafpath
+from vault.dialogs import downloadFile
+from mapGenerator.mapgenProcess import MapGeneratorProcess
+from fa.maps import getUserMapsFolder
+
+logger = logging.getLogger(__name__)
+
+class MapGeneratorManager(object):
+    def __init__(self):
+        self.releaseUrl = "https://github.com/FAForever/Neroxis-Map-Generator/releases/download/{}/NeroxisGen_{}.jar"
+        self.latestRelease = "https://github.com/FAForever/Neroxis-Map-Generator/releases/latest"
+        self.latestVersion = None
+        self.response = None
+
+        self.currentVersion = Settings.get('mapGenerator/version', "0", str)
+        self.previousMaps = Settings.get('mapGenerator/deleteMaps')
+        self.generatorPath = os.path.join(fafpath.get_libdir(), "map-generator")
+        self.mapsFolder = getUserMapsFolder()
+
+        if self.previousMaps:
+            self.deletePreviousMaps()
+
+    def generateMap(self, mapname):
+        dataString =  mapname.split("neroxis_map_generator",1)[1]
+        version = re.search(r'\_(.*?)\_',dataString).group(1)
+        seed = dataString.rsplit('_', 1)[1]
+
+        actualPath = self.versionController(version)
+
+        if actualPath:
+            auto = Settings.get('mapGenerator/autostart', default=False, type=bool)
+            if not auto:
+                msgbox = QtWidgets.QMessageBox()
+                msgbox.setWindowTitle("Generate map")
+                msgbox.setText("Seems that you don't have the map used this game. Do you want to generate it?<br/><b>" + mapname + "</b>")
+                msgbox.setInformativeText("Map generation is CPU intensive task and may take some time.")
+                msgbox.setStandardButtons(QtWidgets.QMessageBox.Yes | QtWidgets.QMessageBox.YesToAll | QtWidgets.QMessageBox.No)
+                result = msgbox.exec_()
+                if result == QtWidgets.QMessageBox.No:
+                    return False
+                elif result == QtWidgets.QMessageBox.YesToAll:
+                    Settings.set('mapGenerator/autostart', True)
+            
+            # Start generator with progress bar      
+            self.generatorProcess = MapGeneratorProcess(actualPath, self.mapsFolder, seed, version, mapname)
+            
+            # Check if map exists or generator failed
+            if os.path.isdir(os.path.join(self.mapsFolder, mapname)):
+                if self.previousMaps:
+                    self.previousMaps = self.previousMaps + ";" + mapname
+                else:
+                    self.previousMaps = mapname
+                Settings.set('mapGenerator/deleteMaps', self.previousMaps)
+                
+                return mapname
+            else:
+                return False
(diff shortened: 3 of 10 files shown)


src/fa/check.py
```diff
@@ -24,7 +24,13 @@ def map_(mapname, force=False, silent=False):
     if fa.maps.isMapAvailable(mapname):
         logger.info("Map is available.")
         return True
-
+ 
+    if "neroxis_map_generator" in mapname:
```
Geosearchef: Please check if map is generated by matching against a regex. See java client MapGeneratorServiceImpl for that regex.

src/fa/check.py
```diff
@@ -24,7 +24,13 @@ def map_(mapname, force=False, silent=False):
...
+        import client # FIXME
```
Geosearchef: fix

src/fa/maps.py
```diff
@@ -238,7 +238,13 @@ def __exportPreviewFromMap(mapname, positions=None):
         return previews
 
     mapname = os.path.basename(mapdir).lower()
-    mapfilename = os.path.join(mapdir, mapname.split(".")[0]+".scmap")
+    
+    # All these mapgen checks in name here and there
+    # but what can we do? :)
+    if "neroxis_map_generator" in mapname:
```
Geosearchef: write an `isGeneratedMap(name)` function in MapGenManager, check against regex there.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+from vault.dialogs import downloadFile
+from mapGenerator.mapgenProcess import MapGeneratorProcess
+from fa.maps import getUserMapsFolder
+
+logger = logging.getLogger(__name__)
+
+class MapGeneratorManager(object):
+    def __init__(self):
+        self.releaseUrl = "https://github.com/FAForever/Neroxis-Map-Generator/releases/download/{}/NeroxisGen_{}.jar"
+        self.latestRelease = "https://github.com/FAForever/Neroxis-Map-Generator/releases/latest"
```
Geosearchef: is there no way to define those as constant, non object/instance, variables?

Geosearchef: also, don't duplicate the address, derive both of them by combining them with a third base address.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        self.latestVersion = None
+        self.response = None
+
+        self.currentVersion = Settings.get('mapGenerator/version', "0", str)
```
Geosearchef, line 29: is that `"0"` the default?

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        self.previousMaps = Settings.get('mapGenerator/deleteMaps')
+        self.generatorPath = os.path.join(fafpath.get_libdir(), "map-generator")
+        self.mapsFolder = getUserMapsFolder()
+
+        if self.previousMaps:
+            self.deletePreviousMaps()
+
+    def generateMap(self, mapname):
+        dataString =  mapname.split("neroxis_map_generator",1)[1]
+        version = re.search(r'\_(.*?)\_',dataString).group(1)
```
Geosearchef: please use a full map name regex pattern (see java client) and match against the entire map name.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        seed = dataString.rsplit('_', 1)[1]
+
+        actualPath = self.versionController(version)
+
+        if actualPath:
+            auto = Settings.get('mapGenerator/autostart', default=False, type=bool)
+            if not auto:
+                msgbox = QtWidgets.QMessageBox()
+                msgbox.setWindowTitle("Generate map")
+                msgbox.setText("Seems that you don't have the map used this game. Do you want to generate it?<br/><b>" + mapname + "</b>")
```
Geosearchef: It looks like you don't have the map being used by this lobby. Do you want to generate it? (also, no internationalization in the py client?)

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+                msgbox.setInformativeText("Map generation is CPU intensive task and may take some time.")
```
Geosearchef: Map generation is `a` CPU intensive task

Geosearchef: Addtionally, is this really necessary? I think just informing the user that this may take some time or e.g. "Please allow up to 60 seconds" should be enough.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+            # Start generator with progress bar      
+            self.generatorProcess = MapGeneratorProcess(actualPath, self.mapsFolder, seed, version, mapname)
+            
+            # Check if map exists or generator failed
+            if os.path.isdir(os.path.join(self.mapsFolder, mapname)):
+                if self.previousMaps:
+                    self.previousMaps = self.previousMaps + ";" + mapname
+                else:
+                    self.previousMaps = mapname
+                Settings.set('mapGenerator/deleteMaps', self.previousMaps)
```
Geosearchef: I suggest a better self explaining variable name (like mapsToDelete)

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        if self.currentVersion == "0" or not self.latestVersion:
+            self.checkUpdates()
+
+            if self.latestVersion and self.versionController(self.latestVersion):
+                self.currentVersion = self.latestVersion   # mapgen is up-to-date
+                Settings.set('mapGenerator/version', self.currentVersion)
+            elif self.currentVersion == "0":               # if not "0", use older version
+                return False                               # otherwise we don't have any generator at all
+
+        seed = ''.join(["%s" % random.randint(0, 9) for num in range(0, 19)])
```
Geosearchef: This is a bad way to generate a random string of digits.
Also, the generator expects a 64-bit, signed seed as string.
This may go completely haywire if your 19-decimal seed exceeds the 64-bit range.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        mapName = "neroxis_map_generator_{}_{}".format(self.currentVersion, seed)
```
Geosearchef, line 89: maybe extract generated map name to a constant as it's used multiple times

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+
+        return self.generateMap(mapName)
+   
+    def versionController(self, version):
+        name = "NeroxisGen_{}.jar".format(version)
```
Geosearchef: preferrably extract to constant

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        filePath = self.generatorPath + "\\" + name
```
Geosearchef: NOOOOOOOOOOOOOOOOOOOOOOO!
JUST NO!
Never ever concatinate file paths manually.
First of all, this will break on any non Windows based operating system. Therefore you should use a constant/function provided by your programming language that gives you the OS specific path separator. (which for URLs is `/`, getting mad at you windows!)

Secondly, for this case you should always use a join/resolve function that gives you a file/directory inside your path, in python this is provided by os.path.join() I think.

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        
+        # Check if required version is already in folder
+        if os.path.isdir(self.generatorPath):
+            for infile in os.listdir(self.generatorPath):
+                if infile.lower() == name.lower():
```
Geosearchef, line 100: why are you lowercasing here?

src/mapGenerator/mapgenManager.py
```diff
@@ -0,0 +1,156 @@
...
+        # Download from github if not            
+        url = self.releaseUrl.format(version,version)
+        return downloadFile(url, filePath, name, "map generator", silent = False)
+ 
+    def deletePreviousMaps(self):
+        '''Delete maps that were created at previous session'''
+        try:
+            maps = self.previousMaps.split(";")
+            for map in maps:
+                if "neroxis_map_generator" in map: #just in case
```
Geosearchef: regex

src/mapGenerator/mapgenProcess.py
```diff
@@ -0,0 +1,73 @@
...
+        self._progress.setMaximum(250)
+        self._progress.canceled.connect(self.close)
+        self.progressCounter = 1
+        
+        self.map_generator_process = QProcess()
+        self.map_generator_process.readyReadStandardOutput.connect(self.on_log_ready)
+        self.map_generator_process.readyReadStandardError.connect(self.on_error_ready)
+        self.map_generator_process.finished.connect(self.on_exit)
+        
+        self.java_path = os.path.join(fafpath.get_libdir(), "ice-adapter", "jre", "bin", "java.exe")
```
Geosearchef: duplicated code, provide a unified function shared with the adapter code for obtaining the jre location.

Geosearchef (changes requested):
Didn't check anything python/code quality related.
Also just a quick review.

Please add a timeout that kills the generator. (~60 seconds)

src/fa/check.py
Wesmania: If importing at the top does not cause issues, do it. If it does, leave it as it is - no point trying to fix client singletons when it's on its way out.

src/mapGenerator/mapgenManager.py
Wesmania: Nope.

src/mapGenerator/mapgenManager.py
Wesmania: It's good enough IMO. Just make it generate 8 bytes, then concatenate them as hex and we're good.

src/mapGenerator/mapgenManager.py
Wesmania: +1, use os.path.join.

src/mapGenerator/mapgenManager.py
Wesmania: Does the map generator preserve case everywhere? I imagine this is trying to be on the safe side, since there's multiple maps that mix case between zips, directories and files. Windows doesn't care, linux does.

src/mapGenerator/mapgenManager.py
Strogoo: yep

src/mapGenerator/mapgenManager.py
Strogoo: Yeah, I know my engrish skills are epic, that's why I always try to copy important msgs from somewhere https://github.com/FAForever/client/blob/develop/src/fa/check.py#L35 :D

src/mapGenerator/mapgenManager.py
Strogoo: Mmm, idk. If user has potato pc he might starts crying "I launched map gen and my PC freezes/cpu overheated/died". And then we can answer that he was informed about cpu load :D

Strogoo:
Ok, need advice about where to place isGeneratedMap function.
see commit above. using option 2 there

1. inside Manager object and then use 
```
import client
if path(client.instance):
```
in every place we need it

2. in mapgenUtils and import it normally where needed

(*don't tell me to fix circular imports in whole client pls, at least not now :D)

src/mapGenerator/mapgenManager.py
Geosearchef: That's not how it's supposed to work. Where's the trouble with generating a random 64 bit integer?

Strogoo:
Ok, Dragonite is suggesting to remove useless class from mapgenManager and make it with bunch of functions. I kinda like that idea to not use classes everywhere no matter what =)

src/fa/check.py
norraxx: We have found the way

src/mapGenerator/mapgenProcess.py
Strogoo: hmm, currently it launches ice via faf-ice-adapter.exe from this archive https://github.com/FAForever/java-ice-adapter/releases/download/v1.0.0/faf-ice-adapter-jre-base.7z

anyway, I'll make a function in utils

src/mapGenerator/mapgenManager.py
Strogoo: Ok, I don't know why but when I use 
```random.getrandbits(64)```
or
```int.from_bytes(os.urandom(8), "big")```

they both freeze client on a second attempt (first is ok). While current rng always works smooth and never freezes

Strogoo: NVM, freeze happens when cmd window has selection or smth. Not rng fault

Strogoo:
I left MapGeneratorManager as a class, because converting it in a bunch on functions doesn't solve some import errors on start if isGeneratedMap() is located in mapgenManager.py. Placing it in mapgenUtils is fine.

PR #1111, opened 2021-01-26
Add lite team matchmaking

Gatsik:
* Add lite version of team matchmaker and party system
 * Fix ability to view maps pools

Still not sure if parties work as expected, because probably no one has played yet a single game with it, so no complaints were received.
It has pitiful UI – it's just a label with your teammate's nickname and buttons to kick or leave a party. No internal chat like in the Java client and to button to invite someone (only in chat).


Gatsik:
Probably should've split this into two PRs

Issue #22, opened 2014-09-25
PEP-8 Compliance


thygrrr:
Refactor Code so it's PEP-8 compliant.

Sheeo:
Issue not done :)

duk3luk3:
Related to #459

PR #1114, opened 2021-08-07
Update client

### LINKED ISSUE #867: Client's window is always open
Clicking the client's icon in the taskbar doesn't minimize it.
The only way to do this is via Win-D to shot the desktop. But then if I click on any other open window the client will pop up as well.

Only if I click the (-) icon to minimize it, it works as intended.

Gatsik:
* Login with OAuth tokens
 * Use new server protocol
 * Add advanced login parameters (#1096)
 * Remove some deprecations
 * Replace hardcoded irclib with external library
 * Add map generator support (#1085)
 * Support zstd compression in replay files
 * Add leaderboard from API
 * Rework matchmaking
   - Join party channels (in chat window for now)
   - Prepare for 3v3 and 4v4 matchmaker release
 * Add more options to leaderboard context menu
 * Allow to foe moderators
 * Disable ICE adapter info window by default
 Maybe something else, it is all should be in changelog

 Fixes #867


Wesmania (approved):
lgtm

src/api/player_api.py
```diff
@@ -0,0 +1,40 @@
...
+        self.leaderboardName = leaderboardName
+        self.request(queryDict, self.handleDataForLeaderboard)
+
+    def handleDataForLeaderboard(self, message, meta):
+        preparedData = dict(
+             command = 'stats'
+            ,type = 'player'
+            ,leaderboardName = self.leaderboardName
+            ,values = message
+            ,meta = meta['meta']
```
Askaholic, line 22: Put comma on previous line?

src/api/player_api.py
```diff
@@ -0,0 +1,40 @@
...
+            'filter' : '(login=="{name}",names.name=="{name}")'.format(name=nameToFind),
+            'fields[player]' : 'login,names',
+            'fields[nameRecord]' : 'name,changeTime,player'
+        }
+        self.request(queryDict, self.handleDataForAliasViewer)
+
+    def handleDataForAliasViewer(self, message, meta=None):
+        preparedData = dict(
+             command = 'alias_info'
+            ,values = message
```
Askaholic, line 38: Lol

src/api/stats_api.py
```diff
@@ -0,0 +1,38 @@
...
+    def requestData(self, queryDict={}):
+        self.request(queryDict, self.handleData)
+
+    def handleData(self, message, meta):
+        preparedData = dict(
+             command = 'stats'
+            ,type = 'leaderboardRating'
+            ,leaderboardName = self.leadeboardName
+            ,values = message
+            ,meta = meta['meta']
```
Askaholic, line 20: Here too

src/api/vaults_api.py
```diff
@@ -20,10 +22,13 @@ def handleData(self, message, meta):
                 ,page = meta['meta']['page']
             )
             self.dispatch.dispatch(data)
+        preparedData = dict(
+             command = 'modvault_info'
+            ,values = []
```
Askaholic, line 27: Everywhere

src/fa/replay.py
```diff
@@ -26,14 +27,26 @@ def replay(source, detach=False):
         featured_mod_versions = None
         arg_string = None
         replay_id = None
+        compression_type = None
         # Convert strings to URLs
         if isinstance(source, str):
             if os.path.isfile(source):
                 if source.endswith(".fafreplay"):  # the new way of doing things
-                    replay = open(source, "rt")
+                    replay = open(source, "rb")
```
Askaholic, line 35: Use `with` statement?

src/fa/replayparser.py
```diff
@@ -37,3 +37,14 @@ def getVersion(self):
             return None
         else:
             return supcomVersion.split(".")[-1]
+
+    def getMapName(self):
+        f = open(self.file, 'rb')
```
Askaholic, line 42: Use `with` statement

src/games/automatchframe.py
```diff
@@ -0,0 +1,218 @@
...
+                partial(self.selectFaction, factionID=faction.value)
+            )
+
+    def handleQueueInfo(self, message):
+        for queue in message.get("queues", {}):
+            if queue["queue_name"] == self.queueName:
+                self.labelInQueue.setText(
+                    "In Queue: {}".format(queue["num_players"])
+                )
+                self.secondsToAutomatch = int(queue["queue_pop_time_delta"])
```
Askaholic, line 93: O.o

Askaholic (commented):
Sure, I mean lots of questionable style choices but whatever.

Gatsik:
yes, i was young and stupid writing commas at the beginning of the line. when i was using SQL, it was easier in this way not to forget delete the comma at the previous line while removing line from the request to avoid error. but it turned out that Python ignores it.

anyway there are a lot of other style issues and inconsistencies. for the last month or half a month i've been trying to write PEP compliant code, so when i did https://github.com/FAForever/client/pull/1114/commits/f458b78bdea9e1f3f57a33127383fbe1019b37ed i decided that it would be better to change all at once in a separate PR, moreover there is an issue for that (#22)

src/games/automatchframe.py
Gatsik: sorry, can't understand you. probably it would be better to update all text labels in one function, but i didn't want to update another label every second.
using `int` i think is better, because i would prefer to stare for almost another second at `00:00` than see match found message at `00:01`

Askaholic:
Oh yea that's super annoying in SQL. Made me laugh though, it looks so strange

src/games/automatchframe.py
Askaholic: Oh, it's just I think the java client doesn't even use this field yet since I added it after `queue_pop_time` was sometimes unreliable when people had weird system times. But technically `queue_pop_time_delta` will be a bit off depending on network latency.

PR #1132, opened 2022-08-28
Added cloudfare verification

ghost:
Verification for patch files.
_hardcoded_

### DIFF
--- a/src/downloadManager/__init__.py
+++ b/src/downloadManager/__init__.py
@@ -4,6 +4,7 @@ import urllib.error
 import urllib.parse
 import urllib.request
 
+from urllib.parse import urlparse, parse_qs
 from PyQt5 import QtCore, QtGui, QtWidgets
 from PyQt5.QtCore import QObject, QUrl, pyqtSignal
 from PyQt5.QtNetwork import QNetworkAccessManager, QNetworkRequest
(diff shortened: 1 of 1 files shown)


src/downloadManager/__init__.py
```diff
@@ -69,7 +70,16 @@ def _finish(self):
...
         self._running = True
+        token = Settings.get("oauth/token")['access_token']
+
+        o = urlparse(self.addr)
+        query = parse_qs(o.query)
+
         req = QNetworkRequest(QUrl(self.addr))
+        if 'verify' in query:
+            req = QNetworkRequest(QUrl(o._replace(query=None).geturl()))
+            req.setRawHeader(b'Verify', query['verify'][0].encode())
```
Gatsik, line 81: Unfortunately, this doesn't work for me, because `parse_qs` decodes percent-encoded sequences and the server can't validate this correctly (unless we manually mess with the encoding).
What about using PyQt5's `QUrlQuery` for this? (It [doesn't decode everything](https://doc.qt.io/qt-5/qurlquery.html#handling-of-spaces-and-plus))
something like this:
```Python
qurl = QUrl(self.addr)
query = QUrlQuery(qurl)
qurl.setQuery(QUrlQuery())
req = QNetworkRequest(qurl)
if not query.isEmpty():
    for item_key, item_value in query.queryItems():
        req.setRawHeader(item_key.encode(), item_value.encode())
    req.setRawHeader(b'Authorization', b'Bearer ' + token.encode())
```

Gatsik:
Superseded by https://github.com/FAForever/client/pull/1135

PR #1135, opened 2024-06-04
Make it operational again

Gatsik:
Port to `PyQt6` and `python3.12` with many small fixes here and there
(Qt has stopped including 32-bit packages, meaning we also can only support 64-bit now. Anyway, official client dropped 32-bit way earlier)


Brutus5000:
@Gatsik I don't expect anybody to review this huge change. You have maintainer permissions, feel free to merge and release whenever it feels right.

Gatsik:
> @Gatsik I don't expect anybody to review this huge change. You have maintainer permissions, feel free to merge and release whenever it feels right.

Sure! I didn't expect reviews too, just wanted to review it by myself from the different perspective (interface)

